PR #3, opened 2022-12-20
Setup Android Project with XMTP proto import

nplasterer:
This imports the xmtp kotlin protos that are published here: https://github.com/xmtp/proto/packages/1754440. And adds a basic test to make sure we can consume them in a library module which will become the sdk in the future.
It also sets up an app module so that we can build the app and write the sdk at the same time.

Currently the app looks like this:
Screenshot 2022-12-20 at 2 02 45 PM


build.gradle
```diff
@@ -0,0 +1,10 @@
+buildscript {
+    ext {
+        compose_version = '1.3.2'
+    }
+}// Top-level build file where you can add configuration options common to all sub-projects/modules.
```
elisealix22, line 5:
```suggestion
}
// Top-level build file where you can add configuration options common to all sub-projects/modules.
```

example/build.gradle
```diff
@@ -0,0 +1,64 @@
...
+    id 'org.jetbrains.kotlin.android'
+}
+
+android {
+    namespace 'org.xmtp.android.example'
+    compileSdk 33
+
+    defaultConfig {
+        applicationId "org.xmtp.android.example"
+        minSdk 28
```
elisealix22, line 12: Any reason for 28 min? Looks like it's 77% of adoption which is _p_ good but dropping down to 27 seems to bring coverage up to 85% which is a p big jump. No worries if this is necessary tho!

Screenshot 2022-12-21 at 3 12 22 PM
Screenshot 2022-12-21 at 3 11 17 PM

>Screenshots are from Android Studio when creating a new project and selecting the OS.

library/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+    id 'org.jetbrains.kotlin.android'
+    id 'com.google.protobuf' version '0.9.1'
+}
+
+android {
+    namespace 'org.xmtp.android.library'
+    compileSdk 33
+
+    defaultConfig {
+        minSdk 28
```
elisealix22, line 12: I guess this is where the Android version matters way more! But same Q as above on if it's necessary.

settings.gradle
```diff
@@ -0,0 +1,26 @@
...
+    repositories {
+        gradlePluginPortal()
+        google()
+        mavenCentral()
+    }
+}
+dependencyResolutionManagement {
+    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
+    repositories {
+        google()
```
elisealix22, line 11: Just to double check: I'm assuming this is still necessary even though it's google() and mavenCentral() are included in the block above?

elisealix22 (approved):
The min SDK question is prob the biggest, but this is EPIC! Amazing work 👏

example/build.gradle
nplasterer: I have a memory about compose or material3 not being supported great all the way back but we can re-adjust that whenever we come across it. I can drop to 27 for now. 👍

nplasterer: We could maybe even go lower on our SDK down to 26 and higher for our example app since mostly UI problems on the lower versions.

library/build.gradle
nplasterer: I said this in my comment above but I think we can go even lower here if wanted since the. Higher versions are more material3 and compose issues if I remember correctly.

settings.gradle
nplasterer: This came by default when I initialized the repository. Whenever I remove them it fails to properly gradle.

example/build.gradle
elisealix22: 26 and 27 didn't seem to make as much of a difference fwiw! (It was like 1%) But agree w the approach.

PR #5, opened 2023-01-05
Base Client and Authorization Signing

nplasterer:
This adds the base client and authorization flow tested with unit tests.
In a follow up I'm adding an integration test with contact bundle but in the interest of keeping these PRs smaller and easier to read I'd like to merge this first.


library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -0,0 +1,75 @@
...
+
+                CiphertextOuterClass.Ciphertext.newBuilder().apply {
+                    aes256GcmHkdfSha256Builder.apply {
+                        payload = final.toByteString()
+                        hkdfSalt = salt.toByteString()
+                        gcmNonce = nonceData.toByteString()
+                    }.build()
+                }.build()
+            } catch (e: GeneralSecurityException) {
+                e.printStackTrace()
```
elisealix22, line 44: It's likely worth removing `e.printStackTrace()`s in favor of Log.e(). We also probably want to strip log statements from the published library to avoid leaking any key info during obfuscation. Something like this can be added to the proguard file so they're still visible while in debug mode but not in any consumers code:

```
-assumenosideeffects class android.util.Log {
    public static boolean isLoggable(java.lang.String, int);
    public static int v(...);
    public static int w(...);
    public static int e(...);
    public static int d(...);
    public static int i(...);
}
```

library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -0,0 +1,75 @@
...
+                val keySpec = SecretKeySpec(key, "AES")
+                val gcmSpec = GCMParameterSpec(128, nonceData)
+
+                cipher.init(Cipher.DECRYPT_MODE, keySpec, gcmSpec)
+                if (additionalData.isNotEmpty()) {
+                    cipher.updateAAD(additionalData)
+                }
+                cipher.doFinal(payload)
+            } catch (e: GeneralSecurityException) {
+                e.printStackTrace()
```
elisealix22, line 70: Same here with favoring `Log` so we can tag it & strip it if we want.

library/src/main/java/org/xmtp/android/library/extensions/Date.kt
```diff
@@ -0,0 +1,6 @@
+package org.xmtp.android.library.extensions
+
+import java.util.*
+
+val Date.millisecondsSinceEpoch: Double
+    get() = (System.currentTimeMillis() * 1000).toDouble()
```
elisealix22, line 6: Any reason this is double vs `createdNs` using Longs below?

library/src/main/java/org/xmtp/android/library/messages/PrivateKey.kt
```diff
@@ -0,0 +1,114 @@
...
+    signedPublicKey.keyBytes = bytes
+    return signedPublicKey.build()
+}
+
+
+
+
+
+
+
```
elisealix22, line 114: Nit: Extra spacing

library/src/main/java/org/xmtp/android/library/messages/Topic.kt
```diff
@@ -0,0 +1,28 @@
+package org.xmtp.android.library.messages
+
+sealed class Topic {
+    data class userPrivateStoreKeyBundle(val v1: String) : Topic()
+    data class contact(val v1: String) : Topic()
+    data class userIntro(val v1: String) : Topic()
+    data class userInvite(val v1: String) : Topic()
+    data class directMessageV1(val v1: String, val v2: String) : Topic()
+    data class directMessageV2(val v1: String) : Topic()
```
elisealix22, line 9: Should this be v2 as a param?

library/src/main/java/org/xmtp/android/library/messages/Topic.kt
```diff
@@ -0,0 +1,28 @@
...
+            return when (this) {
+                is userPrivateStoreKeyBundle -> wrap("privatestore-${v1}")
+                is contact -> wrap("contact-${v1}")
+                is userIntro -> wrap("intro-${v1}")
+                is userInvite -> wrap("invite-${v1}")
+                is directMessageV1 -> {
+                    val addresses = listOf(v1, v2).sorted().joinToString(separator = "-")
+                    wrap("dm-${addresses}")
+                }
+                is directMessageV2 -> wrap("m-${v1}")
```
elisealix22, line 22: Same here: "m-${v2}"?

elisealix22 (commented):
This is EPIC @nplasterer !!! Just a few comments/qs.

Also, I think it might be worth setting up [ktlint](https://pinterest.github.io/ktlint/) sooner rather than later so we don't have to worry about that stuff during reviews or coding. Def not for this PR though!

library/src/main/java/org/xmtp/android/library/extensions/Date.kt
nplasterer: I ended up just removing this entirely. This was something I brought over from iOS and now that I look at it it we don't need it in android. currentTimeMillis is already in millis so we don't have to convert it to millis.

library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -0,0 +1,66 @@
...
+        if (secure) {
+            TlsChannelCredentials.create()
+        } else {
+            InsecureChannelCredentials.create()
+        }
+    ).build()
+
+    private val client: MessageApiGrpcKt.MessageApiCoroutineStub =
+        MessageApiGrpcKt.MessageApiCoroutineStub(channel)
+    private var authToken: String = ""
```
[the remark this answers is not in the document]
nplasterer: I think a lot of this is going to change in my follow up PR but making it nullable for now.

library/src/main/java/org/xmtp/android/library/Constants.kt
```diff
@@ -0,0 +1,3 @@
+package org.xmtp.android.library
+
+data class Constants(val version: String = "0.0.0-development")
```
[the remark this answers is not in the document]
nplasterer: Ya at some point I'll probably need to setup some gradle stuff to manage versioning.

nplasterer:
Good call on Klint. I can add that in a follow up later today hopefully.

library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -0,0 +1,66 @@
...
+        authToken = token
+        return authToken as String
+    }
+
+    suspend fun query(topics: List<Topic>): QueryResponse {
+        val request = QueryRequest.newBuilder()
+            .addAllContentTopics(topics.map { it.description }).build()
+
+        val headers = Metadata()
+        headers.put(AUTHORIZATION_HEADER_KEY, "Bearer $authToken")
```
elisealix22, line 50: With the nullable value, I was thinking we could do something like this. I figure it's better to error bc unauth'd then passing a null / blank value.

```suggestion
        authToken?.let { token ->
            headers.put(AUTHORIZATION_HEADER_KEY, "Bearer $token")
        }
 ```

File: library/src/main/java/org/xmtp/android/library/ApiClient.kt [hunk 10]
```diff
@@ -0,0 +1,66 @@
...
+        }
+    ).build()
+
+    private val client: MessageApiGrpcKt.MessageApiCoroutineStub =
+        MessageApiGrpcKt.MessageApiCoroutineStub(channel)
+    private var authToken: String? = null
+
+    fun setAuthToken(token: String): String {
+        authToken = token
+        return authToken as String
```
elisealix22, line 42: Complete nit, but I think you could return `token` to avoid the cast.

elisealix22 (approved):
🚀  here we go! 2 tiny suggestions but looks great otherwise!!

library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val headers = Metadata()
+        headers.put(AUTHORIZATION_HEADER_KEY, "Bearer $authToken")
+        return client.query(request, headers = headers)
+    }
+
+    suspend fun publish(envelopes: List<Envelope>): PublishResponse {
+        val request = PublishRequest.newBuilder().addAllEnvelopes(envelopes).build()
+        val headers = Metadata()
+        headers.put(AUTHORIZATION_HEADER_KEY, "Bearer $authToken")
```
elisealix22, line 57: I meant to add the same suggestion here as above!

PR #6, opened 2023-01-17
Add Ktlint

nplasterer:
Per @elisealix22 suggested here: https://github.com/xmtp/xmtp-android/pull/5#pullrequestreview-1245739323
Per documentation here: https://pinterest.github.io/ktlint/install/integrations/

This adds a new linter task to our CI and lints the existing code.


.github/workflows/test.yml
```diff
@@ -22,3 +22,28 @@ jobs:
         uses: gradle/wrapper-validation-action@v1
       - name: Gradle Run Tests
         run: ./gradlew library:testDebug
+  static-analysis:
```
nplasterer, line 25: I could be down to moving this into a new workflow file called `lint.yml` instead of adding it to test. Thoughts?

elisealix22: Up to you IMO! Both work for setting up the gh rule I believe so just personal pref

elisealix22 (approved):
💚

.github/workflows/test.yml
nplasterer: I decided to move it. Since we will add the example app tests in this file as well.

PR #9, opened 2023-01-23
Add ApiClient with key bundles, shared secrets, and integration testing

nplasterer:
This PR adds a lot of the base work including key bundles, shared secrets, grpc api client...

This also adds a bunch more unit tests and the ability for integration testing along with the github action to run them.

In a follow up I hope to grow the integration testing to be more complex and full featured.


.github/workflows/test.yml
```diff
@@ -12,13 +12,41 @@ env:
...
+          docker ps
+      - name: Start local test server
+        run: docker-compose -p xmtp -f dev/local/docker-compose.yml up -d
+      - name: Gradle Run Unit Tests
         run: ./gradlew library:testDebug
+      - name: Gradle Run Integration Tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          script: ./gradlew connectedCheck
```
nplasterer, line 49: The more I think about this the more I don't think we need this just yet. The actual full integration test on the device will come once I have an example app but for now my integration tests are just running on the local node which doesn't necessarily need the emulator.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    suspend fun ensureUserContactPublished() {
+        address?.let {
+            val contact = getUserContact(peerAddress = it)
+            if (contact != null && keys?.getPublicKeyBundle() == contact.v2.keyBundle) {
+                return
+            }
+        }
+
+        publishUserContact(legacy = true)
```
elisealix22, line 185: Any reason for the legacy contact here? Could we create a v2 instead?

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun needsIntroduction(peerAddress: String): Boolean =
+        hasIntroduced[peerAddress] != true
+
+    fun find(peerAddress: String): ContactBundle? {
+        val knownBundle = knownBundles[peerAddress]
+        if (knownBundle != null) {
+            return knownBundle
+        }
+        val response =
+            runBlocking(Dispatchers.IO) { client.query(topics = listOf(Topic.contact(peerAddress))) }
```
elisealix22, line 27: Ooooh I think I misunderstood what `runBlocking` does above. Looks like it can be called from the main thread? If so, maybe we can specify `Dispatchers.IO` for `Client.create()` as well?

elisealix22 (approved):
🔥 This is epic, @nplasterer ! Approving because all of my notes/Qs are small. Truly appreciate you adding those integration tests as well to make reviewing this easier!

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: This is still what we do in JS. That's going to change soon-ish, I believe.

PR #10, opened 2023-01-29
Conversations, Messages, and Invitations

nplasterer:
This adds support for conversations, messages, invitations, and codecs.

I'd still like to add some more integration tests before merging this but opening it up for review.


library/src/main/java/org/xmtp/android/library/extensions/Date.kt
```diff
@@ -1,6 +0,0 @@
-package org.xmtp.android.library.extensions
-
-import java.util.Date
-
-val Date.millisecondsSinceEpoch: Double
-    get() = (System.currentTimeMillis() * 1000).toDouble()
```
nplasterer, line 6: This is an iOS thing that we don't need in Android.

library/src/main/java/org/xmtp/android/library/messages/InvitationV1.kt
```diff
@@ -38,9 +39,8 @@ class InvitationV1Builder {
 fun InvitationV1.createRandom(context: Invitation.InvitationV1.Context? = null): InvitationV1 {
     val inviteContext = context ?: Invitation.InvitationV1.Context.newBuilder().build()
     val randomBytes = SecureRandom().generateSeed(32)
-    val randomString =
-        Base64.decode(randomBytes, 0).toByteString().toString().replace(Regex("=*$"), "")
-            .replace(Regex("[^A-Za-z0-9]"), "")
+    val randomString = encodeToString(randomBytes, 0).replace(Regex("=*$"), "")
+        .replace(Regex("[^A-Za-z0-9]"), "")
```
nplasterer, line 43: I keep getting this backwards when reading iOS code. But this should be encoding and a test found that 😄

nplasterer:
Half the commits in here where merged in a previous PR. Not sure why they are still showing up even though I rebased on main after it merged.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -181,18 +194,18 @@ class Client() {
...
+    fun ensureUserContactPublished() {
         address?.let {
             val contact = getUserContact(peerAddress = it)
             if (contact != null && keys?.getPublicKeyBundle() == contact.v2.keyBundle) {
                 return
             }
         }
 
-        publishUserContact(legacy = true)
+        runBlocking { publishUserContact(legacy = true) }
```
elisealix22, line 208: Does this need to be in a `runBlocking` if it's not a `suspend` any more? Just asking out of curiosity

library/src/test/java/org/xmtp/android/library/TestHelpers.kt
```diff
@@ -0,0 +1,119 @@
...
+            )
+        }
+        return QueryResponse.newBuilder().also {
+            it.addAllEnvelopes(result)
+        }.build()
+    }
+
+    override suspend fun publish(envelopes: List<MessageApiOuterClass.Envelope>): MessageApiOuterClass.PublishResponse {
+        for (envelope in envelopes) {
+//            send(envelope = envelope)
```
elisealix22, line 101: Is this suppose to be commented out for now?

library/src/main/java/org/xmtp/android/library/ConversationV1.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    fun <T : Any> send(content: T, options: SendOptions? = null) {
+        val codec = Client.codecRegistry.find(options?.contentType)
+
+        fun <Codec : ContentCodec<T>> encode(codec: Codec, content: Any): EncodedContent {
+            val contentType = content as? T
+            if (contentType != null) {
+                return codec.encode(content = contentType)
+            } else {
+                throw java.lang.NullPointerException()
```
elisealix22, line 52: Small suggestion: I see us using `IllegalArgumentException` elsewhere with a custom message -- maybe that would be more clear than a generic NPE?

library/src/main/java/org/xmtp/android/library/ConversationV2.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    fun <T> send(content: T, options: SendOptions? = null) {
+        val codec = Client.codecRegistry.find(options?.contentType)
+
+        fun <Codec : ContentCodec<T>> encode(codec: Codec, content: Any?): EncodedContent {
+            val contentType = content as? T
+            if (contentType != null) {
+                return codec.encode(contentType)
+            } else {
+                throw java.lang.NullPointerException()
```
elisealix22, line 102: Same here with the NPE but just an optional suggestion!

elisealix22 (approved):
Kotlin-wise this looks great to me! The only thing that jumped out to me is that we use a few different types of exceptions that might not be clear to the consumer on what to catch. Maybe it might make sense to make our own XMTPException class that takes a message and use that everywhere so people can always catch that type. That idea is inspired by [ApolloException](https://www.apollographql.com/docs/kotlin/essentials/errors/) from the Apollo Android library and is def not required for this PR!

library/src/test/java/org/xmtp/android/library/TestHelpers.kt
nplasterer: Yes this is stream related but I can just remove it!

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -0,0 +1,102 @@
+package org.xmtp.android.library
+
+import java.util.Date
+
+sealed class ConversationContainer {
```
nakajima, line 5: You might not need this. The container thing was my first attempt at allowing exporting/importing of conversations, but it's a different format than the JS uses. I eventually added the ability to use the JS format for importing in https://github.com/xmtp/xmtp-ios/pull/53, and it might make sense to just use that for import/export in Kotlin instead of the container thing.

nakajima (approved):
I don't super know Kotlin but everything I saw here made sense to me. Good stuff!

library/src/main/java/org/xmtp/android/library/Conversation.kt
nplasterer: Nice catch I wasn't even using this code anyways so removed it!

nplasterer:
Thanks for all the review. I added several more tests and will address the error messages in a follow up! I was in a UX interview where the integrator said they look for useful error messages first whenever they debug so I know this is a big area for me to improve this codebase.

PR #18, opened 2023-02-13
XMTPException 

nplasterer:
This adds a XMTPException to help with handling XMTP errors in clients.
Also fixes a small V1 bug with topics.


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -120,7 +119,6 @@ class Client() {
         }
     }
 
-    @WorkerThread
```
nplasterer, line 123: Had to remove these annotations since they were the only thing that was requiring importing the entire androidx library.

library/src/androidTest/java/org/xmtp/android/library/InstrumentedTest.kt
```diff
@@ -123,91 +97,4 @@ class InstrumentedTest {
...
-            "0xf4BF19Ed562651837bc11ff975472ABd239D35B5",
-            InvitationV1ContextBuilder.buildFromConversation("https://example.com/4")
-        )
-
-        convo.send(content = "hello from kotlin")
-        Thread.sleep(1_000)
-        val messages = convo.messages()
-        assertEquals(1, messages.size)
-        assertEquals("hello from kotlin", messages[0].body)
-    }
```
nplasterer, line 212: I moved these into unit tests since they didn't need to run on a blocking thread.

elisealix22 (approved):
Love! Feels really clean to me

Issue #20, opened 2023-02-14
Feature request: Add pagination to conversation list


nplasterer:
### Describe the uses cases for the feature

Add pagination

PR #25, opened 2023-02-15
Pagination

nplasterer:
Closes https://github.com/xmtp/xmtp-android/issues/20

This adds the ability to paginate conversation lists.
I also added support for importing conversations.


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -209,6 +226,59 @@ class Client() {
...
+            )
+        )
+    }
+
+    fun importV1Conversation(export: ConversationV1Export): Conversation {
+        val sentAt = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            Date.from(Instant.parse(export.createdAt))
+        } else {
+            Date()
+        }
```
nplasterer, line 272: All of these APIs that allow parsing a UTC date string into a date require API level 26 we currently are targeting 22. This isn't the first time I've run into issues and had to do work around to be able to support 22.
In this case if someone is using an API below 26 and import a V1 conversation (rare) all the conversations will be incorrectly dated to the date that they imported and not the original date. I tried a bunch of different things but wasn't able to make anything work for below 26. Open to suggestions.

library/src/androidTest/java/org/xmtp/android/library/InstrumentedTest.kt
```diff
@@ -97,4 +100,65 @@ class InstrumentedTest {
...
+        // Publish alice's contact
+        Client().create(account = alice, clientOptions)
+        val convo = bobClient.conversations.newConversation(
+            alice.address,
+            context = InvitationV1ContextBuilder.buildFromConversation("hi")
+        )
+        // Say this message is sent in the past
+        val date = Date()
+        date.time = date.time - 10000
+        convo.send(text = "10 seconds ago", sentAt = date)
```
elisealix22, line 120: Kotlin's Date should have `minus()` and `plus()` methods to make this a little easier! https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/minus.html

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -57,13 +66,21 @@ class Client() {
     constructor(
         address: String,
         privateKeyBundleV1: PrivateKeyBundleV1,
-        apiClient: ApiClient
+        apiClient: ApiClient,
```
elisealix22, line 69:
```suggestion
        apiClient: ApiClient
```

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -97,7 +114,7 @@ class Client() {
 
     private suspend fun loadOrCreateKeys(
         account: SigningKey,
-        apiClient: ApiClient
+        apiClient: ApiClient,
```
elisealix22, line 117:
```suggestion
        apiClient: ApiClient
```

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -125,7 +142,7 @@ class Client() {
 
     private suspend fun loadPrivateKeys(
         account: SigningKey,
-        apiClient: ApiClient
+        apiClient: ApiClient,
```
elisealix22, line 145:
```suggestion
        apiClient: ApiClient
```

elisealix22 (commented):
Looks good to me other than I do worry about a min version of 26 as I believe some of our partners are targeting 23. I don't have any suggestions off the top of my head, but I can take a look today 👀

elisealix22 (approved):
Approving since the import will only fail for v1 conversations prior to API 26 and we are phasing them out.

library/src/androidTest/java/org/xmtp/android/library/InstrumentedTest.kt
nplasterer: Ya I looked into this but you have to change the date to a instant and then back to a date which felt like a lot for a test.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -209,6 +229,61 @@ class Client() {
...
+        )
+    }
+
+    fun importV1Conversation(export: ConversationV1Export): Conversation {
+        val sentAt = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            Date.from(Instant.parse(export.createdAt))
+        } else {
+            val df = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.getDefault())
+            df.timeZone = TimeZone.getTimeZone("UTC")
+            df.parse(export.createdAt)
```
nplasterer, line 276: @elisealix22 I found something that works with version 22. SimpleDateFormat is not ideal since if anything differently formatted gets returned this will break so I'm keeping these still split into a sdk version check. But now the date will be accurate.

PR #27, opened 2023-02-15
Set up infrastructure for example app

elisealix22:
This PR adds core infrastructure to be able to build our example app:
- `MainActivity`: Will be the authenticated landing page. Currently it displays the signed in address.
- `ConnectWalletActivity`: Will be the surface we handle all wallet connection flows in. Currently it only generates a client with a new random wallet.
- `ConnectWalletViewModel`: Demonstrates how we'll handle async logic impacting UI state. Currently it creates a bg thread to create a client and with a random address. We're using the latest and greatest [StateFlow](https://developer.android.com/kotlin/flow/stateflow-and-sharedflow#stateflow) to collect UI changes and render them in the `ConnectWalletActivity`.

**Note:** This PR also suggests wrapping `Client#create` in a try/catch to bottle it up into an XMTPException after noticing a crash when I had it misconfigured. I'm happy to revert this if you disagree @nplasterer !

https://user-images.githubusercontent.com/556051/219141696-294dc2ef-51c2-46d5-bb49-8900f69722e5.mp4


nplasterer (approved):
Amazing looks great!

PR #31, opened 2023-02-20
Store keys in AccountManager

elisealix22:
This PR generates a random wallet and stores the serialized private key data in Android's [AccountManager](https://developer.android.com/reference/android/accounts/AccountManager). It then reads the private key data from the AccountManager's `password` field and creates a `Client` using that data.

| System Account | Signed in | 
|--|--|
| Screenshot_1676923185 | Screenshot_1676923175 |


nakajima (approved):
I know nothing about android or kotlin, but looks good to me!

nplasterer (approved):
LGTM. Thanks for writing a test with the library changes. 👍

PR #35, opened 2023-02-23
Consume the new Maven Proto

nplasterer:
We are now publishing the proto repo to maven central 🎉 

https://central.sonatype.com/artifact/org.xmtp/proto-kotlin/3.13.4


elisealix22 (approved):
This is absolutely EPIC 🥲 👏

PR #37, opened 2023-02-24
Add conversation detail example

elisealix22:
WIP

### DIFF
--- a/example/src/main/java/org/xmtp/android/example/conversation/ConversationDetailViewModel.kt
+++ b/example/src/main/java/org/xmtp/android/example/conversation/ConversationDetailViewModel.kt
@@ -1,0 +1,52 @@
+package org.xmtp.android.example.conversation
+
+import androidx.annotation.UiThread
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+import org.xmtp.android.library.Conversation
+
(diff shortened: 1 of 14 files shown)


example/src/main/java/org/xmtp/android/example/conversation/ConversationDetailViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    fun fetchMessages(peerAddress: String) {
+        when (val uiState = uiState.value) {
+            is UiState.Success -> _uiState.value = UiState.Loading(uiState.listItems)
+            else -> _uiState.value = UiState.Loading(null)
+        }
+        viewModelScope.launch(Dispatchers.IO) {
+            val listItems = mutableListOf<MessageListItem>()
+            try {
+                conversation?.let {
+                    it.messages().map { message ->
```
elisealix22, line 28: @nplasterer i'm wondering if maybe we can get a single conversation from a client here instead of having to pass a convo from activity to activity 💭 . Either that or I think we'll want a way to make Conversation parcelizable so we can pass it in and call `messages()` on it directly. Will be fun to pair on this together!

library/build.gradle
```diff
@@ -74,7 +74,7 @@ dependencies {
     implementation 'io.grpc:grpc-protobuf:1.51.0'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
     implementation 'org.web3j:crypto:5.0.0'
-    implementation 'org.xmtp:proto-kotlin:3.13.4'
+    api 'org.xmtp:proto-kotlin:3.15.0'
```
nplasterer, line 77: I think this is fine probably no harm in doing this.

nplasterer (approved):
Thanks for writing a test looks good to me!

PR #38, opened 2023-02-27
Add create conversation bottom sheet

elisealix22:
This PR adds a create conversation bottom sheet to test the end-to-end flow of creating and listing conversations.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -67,7 +67,7 @@ dependencies {
     implementation 'io.grpc:grpc-protobuf:1.51.0'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
     implementation 'org.web3j:crypto:5.0.0'
-    implementation 'org.xmtp.proto:kotlin:3.12.3'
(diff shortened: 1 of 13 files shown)


library/build.gradle
```diff
@@ -67,7 +67,7 @@ dependencies {
     implementation 'io.grpc:grpc-protobuf:1.51.0'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
     implementation 'org.web3j:crypto:5.0.0'
-    implementation 'org.xmtp.proto:kotlin:3.12.3'
+    api 'org.xmtp.proto:kotlin:3.12.3'
```
elisealix22, line 70: This exposes the proto library to the example app which was necessary to call `createConversation(peerAddress, invitation) since invitation is a proto class. I think we _could_ get around this by aliasing invitation, but idk if it's necessary! Would love your input.

nplasterer: Hmmm I think we should just alias it. We do a lot of aliasing and I wouldn't want to make other consumers have to be confused by using our proto library since a lot of classes are kinda similar.

PR #41, opened 2023-03-02
Successful Maven Publish

nplasterer:
I tested this javadoc and jar creation with github packages to confirm it will work for maven.

See package here: https://github.com/xmtp/xmtp-android/packages/1797061


nakajima (approved):
Feeling really confident that this is going to work

PR #43, opened 2023-03-02
Don't specify packaging for maven

nplasterer:
We are actually packaging it as an aar so specifying that it was a jar may cause it to break

See handy comment here:
https://issues.sonatype.org/browse/OSSRH-89137


elisealix22 (approved):
🙏

PR #45, opened 2023-03-05
Implement Remote Attachments

nakajima:
I have never written kotlin before so feel free to tear this all up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Used for test:
+library/abcdefg
+
 # Built application files
 *.apk
 *.aar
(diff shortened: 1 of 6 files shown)


nplasterer (approved):
This looks great thanks for doing this. I just made some minor kotlin changes here: https://github.com/xmtp/xmtp-android/pull/45/commits/37b5bf2961720952af8b3c24a0acc6e1758935ef but if that looks good to you everything else looks good to me. 👍  we can cut a release after this lands.

nakajima:
@nplasterer Awesome, thanks so much for that.

PR #46, opened 2023-03-07
Implement streamAllMessages

nplasterer:
This adds the ability to stream all messages in android.
Also adds a stream messages example to the example app.

https://user-images.githubusercontent.com/3522670/223578540-f6a7292e-79e9-4354-8afd-27c70a70ea37.mp4


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -260,4 +287,29 @@ data class Conversations(
...
+        for (conversation in list()) {
+            topics.add(conversation.topic)
+        }
+        client.subscribe(topics).collect { envelope ->
+            val conversation = conversations.firstOrNull { it.topic == envelope.contentTopic }
+            if (conversation != null) {
+                val decoded = conversation.decode(envelope)
+                emit(decoded)
+            } else if (envelope.contentTopic.startsWith("/xmtp/0/invite-")) {
+                conversations.add(fromInvite(envelope = envelope))
```
neekolas, line 306: Will this update the subscription? I'm not familiar enough with Kotlin flows to understand what happens here. 

Part of the requirements are that when a new conversation is started, the subscriptions will start picking up messages from that conversation.

nplasterer: @neekolas in theory yes. However. Flows are pretty hard to test with unit tests. So I'm probably going to have to write this into the example app to really test it.

README.md
```diff
@@ -24,7 +24,7 @@ For a basic demonstration of the core concepts and capabilities of the `xmtp-and
 
 ## Install from Maven Central
 
-You can find the latest package version on [Maven Central](https://central.sonatype.com/artifact/org.xmtp/android/0.0.4)
+You can find the latest package version on [Maven Central](https://central.sonatype.com/artifact/org.xmtp/android/0.0.5/versions)
```
nplasterer, line 27: @elisealix22 I did end up finding a better link!

jhaaaa (approved):
Thank you for this @nplasterer - README LGTM! <3

example/src/main/java/org/xmtp/android/example/conversation/ConversationDetailViewModel.kt
```diff
@@ -56,6 +67,24 @@ class ConversationDetailViewModel(private val savedStateHandle: SavedStateHandle
         }
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val streamMessages: StateFlow<List<MessageListItem>> =
+        stateFlow(viewModelScope, emptyList()) { subscriptionCount ->
+            if (conversation == null) {
+                conversation = ClientManager.client.fetchConversation(conversationTopic)
+            }
+            conversation!!.streamMessages()
```
elisealix22, line 76: I think we can avoid the force unwrap here similar to how we do above. JUST TO BE SAFE lol:

`conversation?.let {
  it.streamMessages()
`

example/src/main/java/org/xmtp/android/example/conversation/ConversationDetailViewModel.kt
```diff
@@ -90,4 +119,33 @@ class ConversationDetailViewModel(private val savedStateHandle: SavedStateHandle
...
+        started: SharingStarted,
+    ): Flow<T> {
+        return started.command(subscriptionCount)
+            .distinctUntilChanged()
+            .flatMapLatest {
+                when (it) {
+                    SharingCommand.START -> this
+                    SharingCommand.STOP,
+                    SharingCommand.STOP_AND_RESET_REPLAY_CACHE,
+                    -> emptyFlow()
```
elisealix22, line 135: This seems like it could use brackets since the arrow is on the next line. Just a little hard to read and understand it's for both cases

elisealix22 (approved):
Amazing! Added a couple small syntax notes but this is huge.

library/src/main/java/org/xmtp/android/library/Conversations.kt
nplasterer: Okay was able to confirm that this works in test. 👍

PR #48, opened 2023-03-08
Add WalletConnect v1 Flow

elisealix22:
This PR adds a working WalletConnect flow using https://github.com/pink-room/walletconnectkit-android. It also serializes the keys and stores them in the AccountManager to be initialized into the client upon sign in.

In order to initialize a `Client` from a WalletConnect signature we created a new `WalletConnectAccount` class that implements `SigningKey` and handles the `personal_sign` signature.

Once signed, in you'll see a list of addresses which you have conversations with and can tap in to see any messages with that account.

<details>
<summary>Signed in screenshot</summary>

Screenshot_20230310-154054

</details>

### DIFF
--- a/example/src/main/java/org/xmtp/android/example/connect/ConnectWalletViewModel.kt
+++ b/example/src/main/java/org/xmtp/android/example/connect/ConnectWalletViewModel.kt
@@ -1,34 +1,56 @@
 package org.xmtp.android.example.connect
 
-import android.util.Log
+import android.app.Application
 import androidx.annotation.UiThread
-import androidx.lifecycle.ViewModel
+import androidx.lifecycle.AndroidViewModel
 import androidx.lifecycle.viewModelScope
-import com.google.gson.GsonBuilder
-import com.trustwallet.walletconnect.WCClient
-import com.trustwallet.walletconnect.exceptions.InvalidSessionException
-import com.trustwallet.walletconnect.models.WCPeerMeta
-import com.trustwallet.walletconnect.models.session.WCSession
-import java.lang.Thread.sleep
-import java.net.URLEncoder
-import java.util.Random
-import java.util.UUID
+import dev.pinkroom.walletconnectkit.WalletConnectKit
+import dev.pinkroom.walletconnectkit.WalletConnectKitConfig
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
-import okhttp3.OkHttpClient
-import org.komputing.khex.extensions.toNoPrefixHexString
+import org.xmtp.android.library.Client
+import org.xmtp.android.library.SigningKey
 import org.xmtp.android.library.XMTPException
 import org.xmtp.android.library.messages.PrivateKeyBuilder
+import org.xmtp.android.library.messages.buildSignature
+import org.xmtp.proto.message.contents.SignatureOuterClass
 
-class ConnectWalletViewModel : ViewModel() {
+class ConnectWalletViewModel(application: Application) : AndroidViewModel(application) {
 
     private val _uiState = MutableStateFlow<ConnectUiState>(ConnectUiState.Unknown)
     val uiState: StateFlow<ConnectUiState> = _uiState
 
-    private val okHttpClient = OkHttpClient.Builder().build()
-    private val gsonBuilder = GsonBuilder()
+    private val walletConnectKitConfig = WalletConnectKitConfig(
+        context = application,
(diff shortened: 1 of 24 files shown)


example/src/main/java/org/xmtp/android/example/connect/ConnectWalletViewModel.kt
```diff
@@ -31,10 +46,57 @@ class ConnectWalletViewModel : ViewModel() {
...
+                    Log.e("###", "SESSION REQ PEER: $peer")
+                }
+                wcClient.onFailure = { t ->
+                    Log.e("###", "FAILURE")
+                }
+                wcClient.connect(session, peerMeta)
+                _uiState.value = ConnectUiState.Connect(wcURI)
+                for (i in 0 .. 30) {
+                    if (wcClient.isConnected) {
+                        Log.e("###", "CONNECTED!")
```
elisealix22, line 82: I think I might have to try a different WC library since we can't seem to query for address or implement signatures here in order to create a `SigningKey`

elisealix22: Love it! Will give it a shot.

library/src/main/java/org/xmtp/android/library/messages/PrivateKey.kt
```diff
@@ -77,7 +78,7 @@ class PrivateKeyBuilder : SigningKey {
     override val address: String
         get() = privateKey.walletAddress
 
-    override fun sign(data: ByteArray): SignatureOuterClass.Signature {
+    override suspend fun sign(data: ByteArray): SignatureOuterClass.Signature {
```
elisealix22, line 81: @nplasterer I think we'll have to make these suspend functions since it takes some async work to create the signature from the wallet connect kit unless you have a better idea!

library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -31,7 +32,8 @@ fun SigningKey.createIdentity(identity: PrivateKeyOuterClass.PrivateKey): Author
     val signatureClass = Signature.newBuilder().build()
     val signatureText = signatureClass.createIdentityText(key = slimKey.toByteArray())
     val digest = signatureClass.ethHash(message = signatureText)
-    val signature = sign(digest)
+    //TODO(elise): Updated to use signatureText so it would be human readable until we can reverse eth hash
+    val signature = runBlocking { sign(signatureText) } ?: throw XMTPException("Illegal signature")
```
elisealix22, line 36: I noticed the signature message looked like random bytes because we `ethHash` it. I think we'll want to expose a way to un`ethHash` the bytearray in order to display a human readable message when signing if we want to use the digest @nplasterer 👀 . What do you think?

example/src/main/java/org/xmtp/android/example/connect/ConnectWalletViewModel.kt
```diff
@@ -31,10 +68,25 @@ class ConnectWalletViewModel : ViewModel() {
         }
     }
 
+    @UiThread
+    fun connectWallet() {
+        viewModelScope.launch(Dispatchers.IO) {
+            _uiState.value = ConnectUiState.Loading
+            try {
+                val wallet = WCAccount(walletConnectKit)
+                val client = Client().create(wallet)
```
elisealix22, line 77: Unfortunately this is failing for me rn but wanted to push something up to get some eyes on it while I debug!

markusbug: How is it failing? Because in the `PrivateKey` implementation of `SigningKey` you are using `eth_sign` afaik, and walletconnect only supports `personal_sign`.

elisealix22: Was able to resolve the issue! It was a null pointer with the signature creation.

library/src/main/java/org/xmtp/android/library/messages/PrivateKey.kt
```diff
@@ -36,10 +36,6 @@ class PrivateKeyBuilder : SigningKey {
         privateKey = key
     }
 
-    constructor(encodedPrivateKeyData: String) {
```
elisealix22, line 39: I removed these in favor of a privateKeyBundleV1 encoder that can be used both in the generate new wallet & connect existing wallet flows: https://github.com/xmtp/xmtp-android/pull/48/files#diff-7c7d7c0b973262f1159dfb738cbef2159416b59800119179a60f65cb705181d6R11

nplasterer (approved):
This looks great! So excited to have wallet connection working. Once this lands I'll cut a update release version with the sdk changes.

PR #52, opened 2023-03-14
Remove ability to query topics

nplasterer:
In a move towards decentralization we need to remove the ability to query multiple topics as we can't guarantee topics will be on the same node.


mkobetic (approved):
This looks good to my ignorant eye. Thank you so much for jumping on this so quickly. ❤️

elisealix22 (approved):

PR #53, opened 2023-03-14
Improve security

nplasterer:
- prevent injecting messages across topics i.e. rejecting any messages that weren't encrypted/signed for this specific topic.
- prevent an attacker (or compromised XMTP server) from sneaking an invalid/spoofed contact bundle into the Contact topic.
- clean up getting public key from private key


nplasterer:
@nakajima some of these fixes might need to happen in iOS as well. FYI

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -29,6 +30,13 @@ data class Contacts(
         for (envelope in response.envelopesList) {
             knownBundles[peerAddress] = ContactBundleBuilder.buildFromEnvelope(envelope)
```
michaelx11, line 31: We can filter in the for loop, so the code can tolerate a few bad bundles and still return the first legit one.

library/src/main/java/org/xmtp/android/library/messages/MessageV2.kt
```diff
@@ -64,10 +64,13 @@ class MessageV2Builder {
             }.build()
 
             if (key.walletAddress != (PublicKeyBuilder.buildFromSignedPublicKey(signed.sender.preKey).walletAddress)) {
-                throw throw XMTPException("Invalid signature")
+                throw XMTPException("Invalid signature")
             }
             val encodedMessage = EncodedContent.parseFrom(signed.payload)
             val header = MessageHeaderV2.parseFrom(message.headerBytes)
+            if (header.topic != topic) {
```
michaelx11, line 71: nice!

michaelx11 (approved):
Thanks for the quick changes! One requested change on contact bundle, otherwise LGTM

library/src/main/java/org/xmtp/android/library/Contacts.kt
nplasterer: Ahh good call. I fixed that in https://github.com/xmtp/xmtp-android/pull/53/commits/b6ead02697418abcddf625a4f1da55b2e8e8573c

nakajima (approved):

PR #58, opened 2023-03-20
Introduce Conversation.prepareMessage

nplasterer:
This gives back a PreparedMessage that has a messageID on it. Clients can use this for optimistic sending purposes. Clients can then call send() on the prepared message to publish it to the network.

See iOS PR here: https://github.com/xmtp/xmtp-ios/pull/80


elisealix22 (approved):
Love all the tests you added for this! 💚

PR #62, opened 2023-03-30
Add a method to convert bundle to V1 

nplasterer:
JS returns a bundle instead of a bundleV1. This adds a convenience method to make that easier.


library/src/main/java/org/xmtp/android/library/messages/PrivateKeyBundleV1.kt
```diff
@@ -17,6 +19,14 @@ class PrivateKeyBundleV1Builder {
         fun encodeData(privateKeyBundleV1: PrivateKeyBundleV1): String {
             return Base64.encodeToString(privateKeyBundleV1.toByteArray(), Base64.NO_WRAP)
         }
+        fun buildFromBundle(bundleBytes: ByteArray) :PrivateKeyBundleV1 {
```
elisealix22, line 22: Small lint suggestion:
```suggestion
        
        fun buildFromBundle(bundleBytes: ByteArray): PrivateKeyBundleV1 {
            val keys = PrivateKeyOuterClass.PrivateKeyBundle.parseFrom(bundleBytes)
            if (keys.hasV1()) {
                return keys.v1
            } else {
                throw XMTPException("No v1 bundle present")
            }
        }
```

elisealix22 (approved):

PR #63, opened 2023-03-31
Codec Readme

nplasterer:
As discussed it would be helpful to have more documentation around using content types and creating content types. This takes a first pass at that.


library/src/main/java/org/xmtp/android/library/codecs/README.md
```diff
@@ -0,0 +1,71 @@
+# Adding specific content types to your application
+
+You can create any content type you want to be used in your application you can read me about creating content types here: [LINK]
+
+We have a few content types backed into the sdk directly for easy consumption (encryption and decryption) across applications.
+
+TextCodec - (always registered by default) allows the passing of plain text.
+AttachmentCodec - allows the passing of attachments
+RemoteAttachmentCodec - allows the passing of remote attachments
```
nplasterer, line 9: I think we should make this a table that we update as we add more SDK supported content types.

library/src/main/java/org/xmtp/android/library/codecs/README.md
```diff
@@ -0,0 +1,82 @@
...
+```kotlin
+val newConversation = client.conversations.newConversation(walletAddress)
+
+newConversation.send(
+    content = remoteAttachment,
+    options = SendOptions(contentType = ContentTypeRemoteAttachment),
+)
+```
+
+### Receive a remote attachment
```
jhaaaa, line 65: In this step, are we also decrypting and decoding? =)

nplasterer: Yes!

nplasterer:
@nakajima if you get a moment can you read over this doc and make sure this all makes sense?

nakajima (commented):
Everything makes sense to me here.

jhaaaa (approved):
Thank you for this @nplasterer!

PR #64, opened 2023-04-04
feat: expose authCheck etc to enable RN init flow

dmccartney:
This exposes some of the auth flows to support piece-wise authentication in the react-native wrapper.
See https://github.com/xmtp/xmtp-react-native/pull/9

### DIFF
--- a/example/src/main/java/org/xmtp/android/example/pushnotifications/PushNotificationTokenManager.kt
+++ b/example/src/main/java/org/xmtp/android/example/pushnotifications/PushNotificationTokenManager.kt
@@ -37,18 +37,22 @@ object PushNotificationTokenManager {
     }
 
     fun ensurePushTokenIsConfigured() {
-        FirebaseMessaging.getInstance().token.addOnCompleteListener(OnCompleteListener { request ->
-            if (!request.isSuccessful) {
-                Log.e(TAG, "Firebase getInstanceId() failed", request.exception)
-                return@OnCompleteListener
-            }
-            request.result?.let {
-                if (xmtpPushState.value is XMTPPushState.Ready) {
-                    xmtpPush.register(it)
-                    configureNotificationChannels()
+        try {
+            FirebaseMessaging.getInstance().token.addOnCompleteListener(OnCompleteListener { request ->
+                if (!request.isSuccessful) {
+                    Log.e(TAG, "Firebase getInstanceId() failed", request.exception)
+                    return@OnCompleteListener
                 }
-            }
-        })
+                request.result?.let {
+                    if (xmtpPushState.value is XMTPPushState.Ready) {
+                        xmtpPush.register(it)
+                        configureNotificationChannels()
+                    }
+                }
+            })
+        } catch (e: Exception) {
+            Log.e(TAG, "Firebase not setup", e)
+        }
     }
 
     internal fun syncPushNotificationsToken(token: String) {
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -1,6 +1,7 @@
 package org.xmtp.android.library
 
 import android.os.Build
+import android.util.Log
 import com.google.crypto.tink.subtle.Base64
 import com.google.gson.GsonBuilder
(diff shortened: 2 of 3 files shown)


example/src/main/java/org/xmtp/android/example/pushnotifications/PushNotificationTokenManager.kt
```diff
@@ -37,18 +37,22 @@ object PushNotificationTokenManager {
...
+                        xmtpPush.register(it)
+                        configureNotificationChannels()
+                    }
                 }
-            }
-        })
+            })
+        } catch (e: Exception) {
+            Log.e(TAG, "Firebase not setup", e)
+        }
```
nplasterer, line 55: The example app was broken so I pushed up a small fix that will allow people to more easily build the example app without notifications.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -61,6 +62,49 @@ class Client() {
...
+         * The user will need to be prompted to sign to decrypt each bundle.
+         */
+        suspend fun authCheck(api: ApiClient, address: String): List<EncryptedPrivateKeyBundle> {
+            val topic: Topic = Topic.userPrivateStoreKeyBundle(toChecksumAddress(address))
+            val res = api.queryTopic(topic)
+            return res.envelopesList.mapNotNull {
+                try {
+                    EncryptedPrivateKeyBundle.parseFrom(it.message)
+                } catch (e: Exception) {
+                    print("discarding malformed private key bundle: ${e.message}")
```
nplasterer, line 78:
```suggestion
                    Log.d("Client", "discarding malformed private key bundle: ${e.message}", e)
```

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -61,6 +62,49 @@ class Client() {
...
+            api.setAuthToken(authToken)
+            api.publish(
+                envelopes = listOf(
+                    EnvelopeBuilder.buildFromTopic(
+                        topic = Topic.userPrivateStoreKeyBundle(keys.v1.walletAddress),
+                        timestamp = Date(),
+                        message = encryptedKeys.toByteArray()
+                    )
+                )
+            )
```
nplasterer, line 106: Feels like there could be some optimization here instead of calling `keys.v1` 3 different times. Maybe authSave could take v1Keys instead of a PrivateKeyBundle.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -61,6 +62,49 @@ class Client() {
...
+        }
+
+        /**
+         * Use the {@param api} to save the {@param encryptedKeys} for {@param address}.
+         *
+         * The {@param keys} are used to authorize the publish request.
+         */
+        suspend fun authSave(
+            api: ApiClient,
+            keys: PrivateKeyBundle,
```
nplasterer, line 91:
```suggestion
            keys: PrivateKeyBundleV1,
```

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -61,6 +62,49 @@ class Client() {
...
+            encryptedKeys: EncryptedPrivateKeyBundle,
+        ) {
+            val authorizedIdentity = AuthorizedIdentity(keys.v1)
+            authorizedIdentity.address = keys.v1.walletAddress
```
nplasterer, line 95: It's possible that `keys.v1` doesn't exist. A key bundle could include only a v2 key. We should protect against that by throwing an error if the key bundle only includes a v2. I do that in the code here:  https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/messages/PrivateKeyBundleV1.kt#L24-L29

nplasterer (approved):
This looks fine and all the tests are passing and I spun up the example app as well. My only suggestion is around the PrivateKeyBundle versus the PrivateKeyBundleV1. It's been a point of confusion a couple o times before it would be good to make it more clear that you have to have a v1 bundle to be able to create a client. I think the easiest solution would be to just change the param type to a privatekeybundlev1.

PR #66, opened 2023-04-15
Move to lite sdks for protobufs

nplasterer:
This transitions our sdk dependencies to the lite versions to avoid conflicts with other android applications.


library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -77,16 +78,22 @@ data class GRPCApiClient(override val environment: XMTPEnvironment, val secure:
                 }
                 if (pagination?.startTime != null) {
                     it.endTimeNs = pagination.startTime.time * 1_000_000
-                    it.pagingInfoBuilder.direction =
```
nplasterer, line 80: The move to lite mostly just got rid of our fancy builder methods which just makes the code a little more verbose. Hopefully with the move to more kotlin protobufs we can clean a lot of this builder code up.

neekolas (approved):
Seems like if this was going to break something, it would happen at build time or test time. LGTM

PR #67, opened 2023-04-15
[DO NOT MERGE] Add proto files

neekolas:
## Summary

Adds a buf.gen.yaml and generates the proto files. Don't know where we actually want to put them, so I just dumped them in a `proto` folder. Feel free to move the files to a more thoughtful place.


neekolas:
Closing, since it's way out of date. But if someone wants to take this idea and run with it I think it'd be great to get Kotlin codegen out of `xmtp/proto`

PR #75, opened 2023-05-22
Improve sharedSecret performance backed by Rust

nplasterer:
Ran the benchmark on the example app and got `Loaded 2001 conversations in 32468ms`
When run in reactNative here https://github.com/xmtp/xmtp-react-native/pull/40 I got `Loaded 2001 conversations in 28321ms (28.321s)`

I was able to hunt down the slowness to the sharedSecret cryptography code. which is here https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/messages/PrivateKeyBundleV2.kt#L47-L51 
We call this 3 times per conversation so it really adds up as the conversation list grows. In iOS it is backed by the rust library code here: https://github.com/xmtp/libxmtp/blob/main/xmtp_crypto/src/k256_helper.rs#L14:L14

We were able to pull out this costly cryptography code into Rust here: https://github.com/xmtp/libxmtp/pull/125
And now the load time is
` Loaded 2001 conversations in 17040ms`


nakajima (approved):
I can't super speak to the Rust/Kotlin of it all, but it makes sense to me.

PR #77, opened 2023-05-31
Batch query

nplasterer:
Adds the batch query method which will allow for query multiple topics. 

Part of the effort happening here: https://github.com/xmtp/xmtp-react-native/issues/34


library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -126,6 +129,46 @@ data class GRPCApiClient(override val environment: XMTPEnvironment, val secure:
         return query(topic.description, pagination)
     }
 
+    override suspend fun batchQuery(
+        topics: List<String>,
+        pagination: Pagination?,
+    ): BatchQueryResponse {
```
dmccartney, line 135: There’s a little ambiguity in the `pagination` treatment: they get applied per-query-request so the batch endpoint won’t globally sort/limit etc. So if we're going to take a global pagination parameter we should probably implement a merge/sort client side on the results across the batch responses. Or for maximum flexibility you could just make this take a list of QueryRequests.
Here’s where we wrestled with this in flutter land:
https://github.com/xmtp/xmtp-flutter/pull/63#discussion_r1154864940

Also the network will reject requests for >50 `maxQueryRequestsPerBatch`. Which we could either just doc this or we could partition the list of query-requests (or list of topics) into batches of 50 calls.
https://github.com/xmtp/xmtp-flutter/blob/main/lib/src/conversation/conversation_v2.dart#L227-L240

tl/dr:
less prescriptive: take list of query-requests, doc 50 limit
more prescriptive: take list of topics, auto-partition to 50-sized batches, merge and apply pagination to the results

dmccartney (approved):

PR #80, opened 2023-06-07
feat: mem cache convo list, discard decode failures

dmccartney:
This adds memory caching to the `conversations.list()` method -- after the first call, later calls will only look for conversations _newer_ than the most recently listed.

This also changes the message listing and streaming calls to discard malformed messages (instead of throwing an exception and failing to list anything).


nplasterer (approved):
LGMT 👍

PR #97, opened 2023-07-26
Reaction Content Type

nplasterer:
Closes https://github.com/xmtp/xmtp-android/issues/93

This adds the reaction content type to kotlin allowing decryption of reaction messages in push notifications.


library/src/main/java/org/xmtp/android/library/codecs/GroupChatMemberAddedCodec.kt
```diff
@@ -1,11 +1,7 @@
-package uniffi.xmtp_dh.org.xmtp.android.library.codecs
```
nplasterer, line 1: These some how got added under the Rust package. Most of these package changes are just a clean up item.

library/src/main/java/org/xmtp/android/library/codecs/ReactionCodec.kt
```diff
@@ -0,0 +1,43 @@
...
+    val schema: ReactionSchema,
+)
+
+enum class ReactionAction {
+    ADDED, REMOVED
+}
+
+enum class ReactionSchema {
+    UNICODE, SHORTCODE, CUSTOM
+}
```
rygine, line 26: a brief googling on kotlin enum classes says that these will yield UPPERCASE strings. the TypeScript types for `action` and `schema` expect lowercase values. sorry if my understanding of this is wrong.

nplasterer: Yep totally. I just changed them to lowercase.

daria-github (approved):
LGTM!

Issue #102, opened 2023-08-03
Bug: Conversation messages loading ignores limit


gvoltr:
### Describe the bug

Fetching the messages from conversation ignores `limit` argument and always load full conversation.

### Expected behavior

Limit argument affect the result of `conversation.messages()`. Function fetch and return list of messages with size <= limit.

### Steps to reproduce the bug

Request a single message from conversation with multiple messages.

```
val messages = conversation.messages(limit = 1)
print("messages: ${messages.size}")
```

PR #103, opened 2023-08-03
Respect limit in message pagination

nplasterer:
Fixes https://github.com/xmtp/xmtp-android/issues/102

Respects limits for envelope listing.


nakajima (approved):

dmccartney (approved):
LG as fix for #102 

But I think there's a lurking usage ambiguity here. There are two usages of `.messages()`:
- `listAllMessages({... sort parameters ...})` -> will aggregate and yield all pages for you
- `listPaginatedMessages({... sort + limit parameters...}}` -> returns 1 page at a time

Right now `.messages()` is doing both based on whether `limit` is provided. So instead of `limit` having a default value (of, say, 100) instead it has a default value of "invoke possibly massive page aggregation". Users might be less surprised if that "possibly massive" behavior only came from a method like `listAllMessages` and when they're being deliberate they could call and manage `listPaginatedMessages`. 

(meta: we may be able to punt on resolving that ^ since the actual best-practice is to never call `.messages()` and instead use batch listings into a db storage?)

nplasterer:
@dmccartney ya makes sense. But this is how javascript does it. So if we are going to move to a more verbose function naming we should probably change it across all the sdks. We can iterate if people have feedback or concerns.

PR #119, opened 2023-09-18
Return messageId for all conversation.send overloads

gvoltr:
At the moment only `EncodedContent` version of `conversation.send` returns the messageId as result and other send functions are not.

Updated send function result to be consistent across all overloads.


nplasterer (approved):
LGTM!

PR #122, opened 2023-09-28
Subscribe to new conversations in stream all messages

nplasterer:
If I use streamAllMessages , then create a new conversations, messages from the new conversation don’t get streamed.


richardhuaaa (approved):
Nice!! Also tagging @neekolas in case this is relevant to the work he was doing on subscriptions. There's a small risk of losing messages in between closing the old one and starting the new one, but from what I can remember this was the only way to do it pending future work.

PR #131, opened 2023-11-03
Updates to Stream All Messages

nplasterer:
This adds the new subscribe2 function to improve the `streamAllMessages` work.


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -388,12 +392,14 @@ data class Conversations(
...
                             topics.add(conversation.topic)
-                            currentCoroutineContext().job.cancel()
+                            subscribeFlow.value = makeSubscribeRequest(topics)
                         }
 
                         else -> {}
                     }
                 }
+            } catch (error: CancellationException) {
+                break
```
nplasterer, line 402: Only break if it's a cancellation request for the stream? Or should we break for all exceptions?

nakajima (approved):

PR #132, opened 2023-11-03
fix use a local cache link in README.md

Babak-gh:
There was a wrong link in README.md file in "Use local storage" section. I tried to find the correct link in xmtp.org documentations and fixed the link.


nplasterer (approved):
Thanks so much for the fix!

PR #138, opened 2023-11-22
Removing conversations that contain a no valid topic

giovasdistillery:
### Expectations
- Adding a filter to not show conversations that contain no valid topics
- The filter for now is allowing alphanumeric characters and `-` and `/`

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -175,13 +175,22 @@ data class Conversations(
             }
         }
 
-        val regex = Regex("[A-Za-z0-9\\-\\/]+")
         conversationsByTopic += newConversations.filter {
-            it.peerAddress != client.address && it.topic.matches(regex)
+            it.peerAddress != client.address && isValidTopic(it.topic)
         }.map { Pair(it.topic, it) }
 
         // TODO(perf): use DB to persist + sort
         return conversationsByTopic.values.sortedByDescending { it.createdAt }
+    }
+
+    fun isValidTopic(topic: String): Boolean {
+        val regex = Regex("^[\\x00-\\x7F]+$")
+        val index = topic.indexOf("0/")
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -175,8 +175,10 @@ data class Conversations(
             }
         }
 
-        conversationsByTopic += newConversations.filter { it.peerAddress != client.address }
-            .map { Pair(it.topic, it) }
+        val regex = Regex("[A-Za-z0-9\\-\\/]+")
```
nplasterer, line 178: So I dug a little bit deeper on this and the conversation topic name is a 32-byte random alphanumeric string, such as `m-AhUFyewKxM1pxr1JmJJkDK3tDn3gQ0Op` We [wrap](https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/messages/Topic.kt#L29) the topic with the xmtp identifying information so that's highly controlled and should never have garbage characters. You can see a [PR](https://github.com/xmtp/xmtp-js/pull/385) that fixed the underlying issue in JS and why we have to filter now because of random conversations that got corrupted. I think if we are going to match this work the Regex would be `/^[\x00-\x7F]+$/`. But this feels like the correct spot.

It also feels worth it to write a [test](https://github.com/xmtp/xmtp-android/blob/main/library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt) to validate that this works. Might be hard to create a conversation with a corrupt topic at this point but you could try and do some manual overrides to confirm and not merge the test if that makes sense.

nplasterer (approved):
On your branch I got numbers around:

`Loaded 2002 conversations in 17306ms`

On main I got numbers around:

`Loaded 2002 conversations in 17199ms`

This seems to validate no performance impact. Thanks for adding the tests. LGTM!

PR #141, opened 2023-11-28
Removing conversations that contain a no valid topic

giovasdistillery:
### Expectations
- Adding a filter to not show conversations that contain no valid topics
- The filter for now is allowing alphanumeric characters and ASCII characters

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/messages/Topic.kt
+++ b/library/src/main/java/org/xmtp/android/library/messages/Topic.kt
@@ -21,10 +21,23 @@ sealed class Topic {
                     addresses.sort()
                     wrap("dm-${addresses.joinToString(separator = "-")}")
                 }
+
                 is directMessageV2 -> wrap("m-$addresses")
                 is preferenceList -> wrap("pppp-$identifier")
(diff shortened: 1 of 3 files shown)


library/src/main/java/org/xmtp/android/library/messages/Topic.kt
```diff
@@ -21,10 +21,23 @@ sealed class Topic {
...
                 is directMessageV2 -> wrap("m-$addresses")
                 is preferenceList -> wrap("pppp-$identifier")
             }
         }
 
     private fun wrap(value: String): String = "/xmtp/0/$value/proto"
+
+    companion object {
+        fun isValidTopic(topic: String): Boolean {
+            val regex = Regex("^[\\x00-\\x7F]+$") // Use this regex to filter non ASCII chars
```
rygine, line 34: i know this function wasn't modified here, but if i'm reading this right, this regex will allow ASCII characters 0-127. characters 0-31 are control characters, and 32 is a space. i don't think we want these.

nplasterer (approved):

PR #144, opened 2023-12-05
feat: add PreEventCallback

kele-leanes:
**Description:** 
This PR introduces the `preEnableIdentityCallback` and `preCreateIdentityCallback` to the ClientOptions object to be called before enabling and creating an identity.
This feature is the starting point of this [issue](https://github.com/xmtp/xmtp-react-native/issues/109)


library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
```diff
@@ -85,6 +89,7 @@ class ClientTest {
     }
 
     @Test
+    @Ignore("CI Issues")
```
giovasdistillery, line 92: I'm not sure why is using this `@Ignore`

giovasdistillery (approved):
LGTM, I just want to know why the use of `@Ignore("CI issues")`

library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
kele-leanes: as you can see [here](https://github.com/xmtp/xmtp-android/actions/runs/7103203135/job/19335347623#step:9:441) we can not run a local environment on CI. is the reason why most of the test that requires create a Client are ignored

nplasterer: You actually can run in the local environment in CI. IOS is the only repo who has that issue. You can see more of how this is setup here https://github.com/xmtp/xmtp-android/blob/main/.github/workflows/test.yml#L51

nplasterer:
Do you mind writing up a detailed issue body so future developers can understand the changes made here?

library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
kele-leanes: so why was the test pass in my local and it didn’t in CI? 🤔

nplasterer: @kele-leanes can you link me to the CI that failed?

kele-leanes: https://github.com/xmtp/xmtp-android/actions/runs/7106161017/job/19344969828 It seems there is an error creating the client

kele-leanes:
@nplasterer 
The purpose of this pull request is to add the optional methods `preCreateIdentityCallback` and `preEnableIdentityCallback`, which are received through the `ClientOptions` and are invoked in the `create` method. These methods are executed just before creating an identity in the case of `preCreateIdentityCallback` when calling `wallet.createIdentity()`, and in the case of `preEnableIdentityCallback`, they are called both when encrypting or decrypting a `keyBundle`. 
This is necessary for integration with React Native, as on the JavaScript side, we may be listening to events (exposed [here](https://github.com/xmtp/xmtp-react-native/pull/177/files#diff-3d1b8496c4f066a0a1d2f776063d85d49f70545d6a94a57fe91eb6efecd50a48R137-R138)) that are triggered once these callbacks are invoked.

nplasterer:
Thanks @kele-leanes for the explanation and working code in RN. I'll give this another pass later today on the tests and get it merged for you.

nplasterer (approved):
Just did a small clean up of some of the methods 👍

PR #153, opened 2024-01-16
fix: Improve performance list method

kele-leanes:
**Description:**
Addressing the concerns raised in [issue #130](https://github.com/xmtp/xmtp-react-native/issues/130), this pull request enhances the pagination functionality for the list method. The key improvement involves introducing a new optional parameter, `createdAt`, which can be set when a conversation is created. This parameter is subsequently utilized in the `importToTopic` method to establish a reference point, enabling the retrieval of the most recent conversations from that specific date. Incorporating this enhancement significantly contributes to optimizing the request time for improved overall performance. 

| 0 conversations imported | 500 conversations imported |
|:--:|:--:|
| Screenshot 2024-01-16 at 18 07 58 | Screenshot 2024-01-16 at 18 05 37 |


library/src/main/java/org/xmtp/android/library/ConversationV2.kt
```diff
@@ -30,6 +30,7 @@ data class ConversationV2(
     val context: Invitation.InvitationV1.Context,
     val peerAddress: String,
     val client: Client,
+    val createdAt: Long? = null,
```
nplasterer, line 33: Same comments as iOS I think createdAt feels more like a date since this is taking NS lets make that more clear

```suggestion
    val createdAtNs: Long? = null,
```

library/src/main/java/org/xmtp/android/library/ConversationV2.kt
```diff
@@ -50,12 +51,13 @@ data class ConversationV2(
...
                 peerAddress = peerAddress,
                 client = client,
+                createdAt = header.createdNs,
                 header = header,
             )
         }
     }
 
-    val createdAt: Date = Date(header.createdNs / 1_000_000)
+    val conversationCreatedAt: Date = Date((createdAt ?: 0L) / 1_000_000)
```
nplasterer, line 60: Lets keep this named the same and fall back to the header if not defined.

```suggestion
    val createdAt: Date = Date((createdAt ?: header.createdNs) / 1_000_000)
```

library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
```diff
@@ -137,4 +138,80 @@ class ConversationsTest {
 
         assertEquals(allMessages.size, 15)
     }
+
+    @Test
+    fun testLoadConvos() {
```
nplasterer, line 143: Again don't want to merge this test just because the potential for variability in the wallet on dev. But appreciate you writing it and testing it.

nplasterer (commented):
Same comments as iOS

nplasterer (approved):

PR #156, opened 2024-01-19
LibXMTP Client Creation

nplasterer:
Part of https://github.com/xmtp/xmtp-android/issues/157

This adds the creation of a V3 client on create of any client. Makes Signing Key extend the FfiInboxOwner so that we can correctly extend and sign. Moved all the group logic to a new branch so that we can land these PRs slowly.


library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -233,6 +254,7 @@ sealed class Conversation {
         return when (this) {
             is V1 -> conversationV1.decryptedMessages(limit, before, after, direction)
             is V2 -> conversationV2.decryptedMessages(limit, before, after, direction)
+            is Group -> TODO()
```
nplasterer, line 257: The decryptedMessage stuff is needed for React Native to be able to support custom content types.

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -184,6 +202,7 @@ sealed class Conversation {
             return when (this) {
                 is V1 -> conversationV1.topic.description
                 is V2 -> conversationV2.topic
+                is Group -> TODO()
```
nplasterer, line 205: Is the id() in V3 the same as a topic? Can we turn the byte array into a string?

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -41,6 +44,7 @@ sealed class Conversation {
             return when (this) {
                 is V1 -> conversationV1.sentAt
                 is V2 -> conversationV2.createdAt
+                is Group -> TODO()
```
nplasterer, line 47: I couldn't find a date on the group is that just not exposed yet or am I missing it?

neekolas: I'll fix that

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -81,6 +83,17 @@ data class Conversations(
         )
     }
 
+    suspend fun newGroup(accountAddress: String): Group {
+        val group = libXMTPConversations?.createGroup(accountAddress)
```
neekolas, line 87: I think it's a mistake in the bindings that `createGroup` takes a single address.

We could either do no address or take an array of addresses (and pre-populate the group for the caller).

nplasterer: I think for usage it would probably make more sense for it to take an array of addresses

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -153,18 +160,38 @@ class Client() {
...
-                val client = Client(account.address, privateKeyBundleV1, apiClient)
+                val libXMTPClient: FfiXmtpClient =
+                    createClient(
+                        logger = logger,
+                        ffiInboxOwner = account,
+                        host = "https://dev.xmtp.network:5556",
+                        isSecure = true,
+                        db = null,
+                        encryptionKey = null
+                    )
```
nplasterer, line 193: Should I be feature flagging this creation? Doesn't seem harmful to do this behind the scenes for all existing clients or is it?

neekolas: The only harm is that it will prompt for a signature in some cases.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -153,18 +160,39 @@ class Client() {
...
+    ): Client {
         return runBlocking {
             try {
                 val privateKeyBundleV1 = loadOrCreateKeys(account, apiClient, options)
-                val client = Client(account.address, privateKeyBundleV1, apiClient)
+                val libXMTPClient: FfiXmtpClient =
+                    createClient(
+                        logger = logger,
+                        ffiInboxOwner = account,
+                        host = "https://dev.xmtp.network:5556",
```
neekolas, line 189: Steven is working on getting MLS endpoints deployed to dev right now, but until then I'd suggest localhost with is_secure set to false.

library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -13,15 +13,16 @@ import org.xmtp.android.library.messages.rawData
...
-interface SigningKey {
-    val address: String
-
+interface SigningKey: FfiInboxOwner {
     suspend fun sign(data: ByteArray): SignatureOuterClass.Signature?
 
-    suspend fun sign(message: String): SignatureOuterClass.Signature?
+    suspend fun signLegacy(message: String): SignatureOuterClass.Signature?
+
+    override fun sign(text: String): ByteArray
```
nplasterer, line 25: Java doesn't allow name shadowing since they both take a String even if they have different return values. To make this change smaller I just renamed the existing sign to `signLegacy` for now but hopefully we can remove it entirely.

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: Also since production won't be available until beta we could wind up with mismatches between dev and prod (prod identity used in dev)

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: I'm getting a `API error: setup error: transport error` now that I changed this to `10.0.2.2:5556` (android equivalent of localhost) Is that the correct endpoint I'm supposed to be pointing at?

nplasterer: O wild it passes but only if I set isSecure to true 🤷‍♀️

neekolas: Interesting. Usually it fails hard connecting to localhost using TLS. Have you tried actually doing anything with the SDK. I don't think it makes any network requests during client instantiation.

Just to confirm, are you using `xmtp/node-go:mls` as your Docker image for the local node?

cameronvoell (dismissed):
Changes look good 👍 

One thought I had was that it might be helpful to add a comment somewhere in the PR or code base marking what commit you're taking the libxmtp bindings from, as it could get confusing with future PRs if changes are being added to libxmtp.

neekolas:
Following up on that. How do we get this to use published releases. 

I just pushed up the [latest as a release here](https://github.com/xmtp/libxmtp/releases/tag/0.4.1)

nplasterer:
@neekolas probably the preferred method would be to publish to maven like we do in the proto repo. But since it's a JNI I'm not sure it would be as straight forward. Definitely something we should prioritize after alpha.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -153,18 +165,47 @@ class Client() {
...
+                        val dbDir = File(options.appContext.filesDir.absolutePath, "xmtp_db")
+                        dbDir.mkdir()
+                        val dbPath: String = dbDir.absolutePath + "/xmtp-${account.getAddress()}.db3"
+                        createClient(
+                            logger = logger,
+                            ffiInboxOwner = account,
+                            host = "http://10.0.2.2:5556",
+                            isSecure = false,
+                            db = dbPath,
+                            encryptionKey = null
```
nplasterer, line 201: @richardhuaaa @neekolas can we move the `dbEncryptionKey` into the rust layer? It feels like something that needs to be derived by the account/privateKey so that it can be easily reproduced anytime client create is called.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -153,18 +165,47 @@ class Client() {
...
+    ): Client {
         return runBlocking {
             try {
                 val privateKeyBundleV1 = loadOrCreateKeys(account, apiClient, options)
-                val client = Client(account.address, privateKeyBundleV1, apiClient)
+                val libXMTPClient: FfiXmtpClient? =
+                    if (options != null && options.enableLibXmtpV3 && options.appContext != null) {
+                        val dbDir = File(options.appContext.filesDir.absolutePath, "xmtp_db")
+                        dbDir.mkdir()
+                        val dbPath: String = dbDir.absolutePath + "/xmtp-${account.getAddress()}.db3"
```
nplasterer, line 194: I'm naming the db after the account address so that there isn't collisions if multiple addresses log in on the same device.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -49,6 +55,8 @@ data class ClientOptions(
     val api: Api = Api(),
     val preCreateIdentityCallback: PreEventCallback? = null,
     val preEnableIdentityCallback: PreEventCallback? = null,
+    val appContext: Context? = null,
+    val enableLibXmtpV3: Boolean = false,
```
nplasterer, line 59: We need the application to pass a context in from the application layer so that we can create a database. Is the `ClientOptions` the correct place for that?
Some for the flag on if V3 is enabled.

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: 👍

neekolas: Maybe `$env-$address` so that you can switch between dev/local/prod in the same app

library/src/main/java/org/xmtp/android/library/Client.kt
richardhuaaa: I would vote in favor of that, curious what @nakajima and @neekolas think.

By the way, I think we will also need a `preGrantMessagingAccess` callback that gets called before signing the text that comes from v3

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -173,6 +201,60 @@ class Client() {
...
+            if (options != null && options.enableLibXmtpV3 && options.appContext != null) {
+                val alias = "xmtp-${options.api.env}-${account.getAddress()}"
+
+                val dbDir = File(appContext?.filesDir?.absolutePath, "xmtp_db")
+                dbDir.mkdir()
+                val dbPath: String = dbDir.absolutePath + "/$alias.db3"
+
+                val sharedPreferences =
+                    options.appContext.getSharedPreferences("XMTPPreferences", Context.MODE_PRIVATE)
+                val dbEncryptionKey = sharedPreferences.getString(alias, null)
```
richardhuaaa, line 220: In the long term, is it possible to put this in the Android Keystore? Shared preferences doesn't appear to be secure from people with access to the device, therefore the encryption key wouldn't provide any additional protection compared to leaving the database unencrypted

library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -50,7 +53,7 @@ fun SigningKey.createIdentity(
     val signatureClass = Signature.newBuilder().build()
     val signatureText = signatureClass.createIdentityText(key = slimKey.toByteArray())
     val digest = signatureClass.ethHash(message = signatureText)
-    val signature = runBlocking { sign(signatureText) } ?: throw XMTPException("Illegal signature")
```
richardhuaaa, line 53: I think we can use the same `suspend` function for `sign` and `signLegacy` now

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -173,6 +201,60 @@ class Client() {
...
+                    //    NETWORK, // v2 keys already on the network
+                    //    KEY_GENERATOR; // I created them
+                    legacyIdentitySource = LegacyIdentitySource.NETWORK,
+                    legacySignedPrivateKeyProto = privateKeyBundleV1.toV2().toByteArray()
+                )
+            } else {
+                null
+            }
+
+        libXMTPClient?.textToSign()?.let {
```
neekolas, line 251: I think I may have led you astray here. You'll need to call register identity one way or the other here. If there is `textToSign` you'll call it with the signature, and if there isn't you'll call it with `null`.

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: So I went down that route here https://github.com/xmtp/xmtp-android/pull/156/commits/1e84e4c34abddd8e9bb4204bee46770d3f2c163e but I wasn't able to make it work as the keystore stuff didn't come till API 23 or 24. I'm going to bump our min sdk version to 23 and see if I can make it work.

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Can you tell me more about this callback/ how I should use it? I'm assuming I can add it later.

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Seems like bumping to version 23 allowed me to use the keystore!

library/src/main/java/org/xmtp/android/library/SigningKey.kt
nplasterer: This was just a rename so to not break existing stuff I'm just going to leave it for now. Probably a larger refactor to tighten up all the sign stuff.

nplasterer:
I think this is getting really close and ready for another round of review.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -137,12 +155,14 @@ class Client() {
         address: String,
         privateKeyBundleV1: PrivateKeyBundleV1,
         apiClient: ApiClient,
+        libXmtpClient: FfiXmtpClient? = null,
     ) : this() {
-        this.address = address
+        this.address = libXMTPClient?.accountAddress() ?: address
```
neekolas, line 160: I'd suggest just using whatever address you already are using for the client address. The one from `libxmtp` is going to be lowercased and the existing one might be checksum cased, so I wouldn't want to cause any V2 issues.

library/src/main/java/org/xmtp/android/library/libxmtp/XMTPLogger.kt
```diff
@@ -0,0 +1,10 @@
+package org.xmtp.android.library.libxmtp
+
+import android.util.Log
+import uniffi.xmtpv3.FfiLogger
+
+class XMTPLogger : FfiLogger {
+    override fun log(level: UInt, levelLabel: String, message: String) {
+        Log.i("$level $levelLabel", message)
```
neekolas, line 8: Would be nice to actually log at different levels by transforming the Rust log levels to the Android log levels.

library/build.gradle
```diff
@@ -14,7 +14,7 @@ dokkaGfmPartial {
 
 ktlint {
     filter {
-        exclude { it.file.path.contains("xmtp_dh") }
```
neekolas, line 17: That's a good reminder. I believe `libxmtp` now has a full set of replacement APIs for what you're using in `xmtp_dh` (if anything is missing we can add it). Maybe an item to add to the backlog for moving over to a single Rust dependency just like iOS did. Obviously not critical.

nplasterer: O good call out. I'm only using 2 rust calls in V2 so I can probably clean those up

library/build.gradle
```diff
@@ -23,7 +23,7 @@ android {
     compileSdk 33
 
     defaultConfig {
-        minSdk 22
+        minSdk 23
```
nplasterer, line 26: Bumping to 23 should be safe as all our integrators are now on it.

library/src/main/java/org/xmtp/android/library/Client.kt
richardhuaaa: Fine to add later. With v3, we're adding a new signature type called 'Grant Messaging Access', it looks like (2) [here](https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/IDENTITY.md#installation-registration), and is what will be passed back from `Client.text_to_sign()`.

`preGrantMessagingAccess` would be the equivalent of `preCreateIdentity`, it gives the app developer a chance to show something in the UI before requesting the signature.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -173,6 +233,72 @@ class Client() {
...
+    private suspend fun ffiXmtpClient(
+        options: ClientOptions?,
+        account: SigningKey?,
+        accountAddress: String,
+        appContext: Context?,
+        privateKeyBundleV1: PrivateKeyBundleV1,
+        legacyIdentitySource: LegacyIdentitySource,
+    ): FfiXmtpClient? {
+        val v3Client: FfiXmtpClient? =
+            if (options != null && options.enableLibXmtpV3 && options.appContext != null) {
```
neekolas, line 245: WDYT about having some validation where if you try and set `enableLibXmtpV3` and have the network set to an environment that doesn't support it we just error out of creation.

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: Maybe we can give this `enableLibXmtpV3` option a friendlier name, if it's going to be visible to integrators. Like `enableGroups`

library/build.gradle
nplasterer: Okay! I was able to remove all the v2 rust stuff 🎉

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: For sure I can get that update to only work when on the local network.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -173,6 +233,72 @@ class Client() {
...
+                )
+            } else {
+                null
+            }
+
+        if (v3Client?.textToSign() == null) {
+            v3Client?.registerIdentity(null)
+        } else {
+            v3Client.textToSign()?.let {
+                v3Client.registerIdentity(account?.sign(it))
```
richardhuaaa, line 295: I think the `account` param of this function should be non-optional - if there is a text to be signed then we must request it, or the v3 client will be non-functional

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -150,21 +170,61 @@ class Client() {
         val clientOptions = options ?: ClientOptions()
         val apiClient =
             GRPCApiClient(environment = clientOptions.api.env, secure = clientOptions.api.isSecure)
-        return Client(address = address, privateKeyBundleV1 = bundle, apiClient = apiClient)
+        val v3Client: FfiXmtpClient? = runBlocking {
+            ffiXmtpClient(
+                options,
+                null,
```
richardhuaaa, line 176: I don't think we can skip the signer here (see other comment)

richardhuaaa (approved):
Well done on the quick turnaround!

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: This case only happens with static keys. So you are passing in a bundle not a signer object. I guess I can make a privateKey from that bundle if it's required 🤔

richardhuaaa: This param is intended to be specifically used for wallet signing, as opposed to signing with the legacy identity key. Even if a legacy identity key was passed in, and it was obtained from a cached/static key, libxmtp may sometimes still need a wallet signature. I'll write a doc later to make all of this clear!

richardhuaaa: Oh - @nplasterer is the issue that the final app integrator won't give us a signer if they give us a cached bundle?

neekolas (approved):

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Correct! Should I go back to the null implementation?

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -150,21 +170,61 @@ class Client() {
         val clientOptions = options ?: ClientOptions()
         val apiClient =
             GRPCApiClient(environment = clientOptions.api.env, secure = clientOptions.api.isSecure)
-        return Client(address = address, privateKeyBundleV1 = bundle, apiClient = apiClient)
+        val v3Client: FfiXmtpClient? = runBlocking {
```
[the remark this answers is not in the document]
nplasterer: Changing all these create methods to suspend would be a breaking change. I'm totally down with doing it if things seem broken but lets hold off till this is out of alpha.

nplasterer:
BREAKING CHANGES introduced in this PR for future release notes.
- minSDKVersion 22 -> 23
- `SigningKey` now takes an addition sign method and the previous sign method has been renamed to `signLegacy`

If you want to try out alpha groups 
- `fromBundle` now requires a `SigningKey`
- all `create` Client functions require `ClientOptions` to `enableAlphaMls` and a `appContext`

PR #158, opened 2024-01-23
Bump to Java version 17

nplasterer:
The latest version of react native require Java 17


nplasterer:
We confirmed on React Native that this shouldn't break existing flows. Other developers may need to update to the latest java to continue using. Fixes https://github.com/xmtp/xmtp-android/issues/167

alexrisch (approved):

Issue #157, opened 2024-01-22
Tracking: Android Group Chat - ALPHA


nplasterer:
### Group Chat Alpha - [Targeting February 1, 2024]

- [ ] Reproduce utility classes/methods
  - [ ] Client
    - [x] update existing `create` method on client to also create a `FfiXmtpClient`
    - [x] Existing client should now also have a `FfiLogger`
    - [x] Have existing SigningKey extend `FfiInboxOwner` so it can be used to create both V2 and V3 clients
    - [x] If `FfiXmtpClient` exists get address from `account_address()`
    - [ ] If `FfiXmtpClient` exists store a reference to `installation_id()`
    - [x] Conversations on client should also take a `libXMTPConversations`
    - [x] If `FfiXmtpClient` exists client create should call `register_identity`
    - [ ] Ability to `list_my_installations()`
    - [ ] Ability to `revoke_installation()`
  - [ ] Conversations
    - [x] Should be able to create a group `newGroup()` -- `create_group()`
    - [x] Should be able to list just groups `listGroups` -- `libXMTPConversations.list()`
    - [x] Should be able to list an ordered list of conversations & groups `list(includeGroups: Boolean = false)`
    - [x] Should be able to stream groups `libXMTPConversations.stream()`
    - [x] Should be able to stream both conversations & groups `stream(includeGroups:Boolean = false)`
    - [x] Should be able to `sync` groups so that the libxmtp database gets updated
    - [x] Should be able to stream all messages from both groups and conversations (`streamAllMessages(includeGroups: Boolean = false)`
  - [ ] Conversation/Group methods
    - [x] Should have a unique identifier of the conversation `id()` equivalent to existing topic
    - [x] Should be able to `send()` messages of any content type
    - [x] Should be able to resync the libxmtp database `sync()`
    - [x] Should be able to list messages `find_messages()` as Decoded and Decrypted Messages
    - [x] Should be able to list members `list_members()`
    - [x] Should be able to add members `add_members()`
    - [x] Should be able to remove members `remove_members()`
    - [x] Should be able to stream messages `stream()`
    - [ ] Should be able to leave a group `leave()`
    - [x] Can check if a message can be sent `can_message()`
- [ ] Publish repo README covering Dev Preview features and with Dev Preview badge

### Docs
[...]

PR #160, opened 2024-01-23
Add Groups to Conversations

nplasterer:
Part of https://github.com/xmtp/xmtp-android/issues/157

This adds a new class called Group that behaves similar to a Conversation and allows listing and sending of messages.


library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    fun testCanCreateAGroup() {
+        val group = boClient.conversations.newGroup(alix.walletAddress)
+        assert(group.id.isNotEmpty())
+    }
+
+    @Test
+    fun testCanListGroupMembers() {
```
neekolas, line 55: @nplasterer I figured out what's happening here. There's an error happening, but it's not bubbling up to you.

The error is casued by the fact that caro and alix haven't had `register_identity` called on them yet, so they don't exist on the network.

Will work on bubbling up these kinds of errors better.

nplasterer: Hmmm I'm pretty sure I am calling register identity. But I added it again and even with explicit register identity. It still doesn't seem to work, for newGroup or for addMember.

neekolas: I pushed up [a commit](https://github.com/xmtp/xmtp-android/pull/160/commits/43290335f78b92d434548cbd82a439bffdf6feec) that fixes a build issue, but I am seeing some errors in the logs I wouldn't expect and the test also isn't passing for me.

The same case is passing in the [libxmtp tests](https://github.com/xmtp/libxmtp/blob/be2afe8226d3a44fe35d6ccd089354e0a0014fa7/bindings_ffi/src/mls.rs#L573), so something must be getting lost in the FFI layer.

nplasterer: Ahh I forgot to push up my local changes that did those things. I'll let you dig into it unless you want a second head. I don't know how helpful I will be debugging the bindings.

neekolas: I found it. It's a casing issue. Once I lowercased the wallet addresses the results started coming back.

We should just standardize all casing at the `libxmtp` level, and none of this should be your problem. I'll put up a PR to do that.

nplasterer: O interesting it couldn't find the identity because I was passing in something that was capitalized! Thanks for looking into it!

neekolas: `libxmtp` will handle the casing for you once https://github.com/xmtp/libxmtp/pull/453 goes in

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -0,0 +1,172 @@
...
+        return encoded
+    }
+
+    suspend fun sync() {
+        libXMTPGroup.sync()
+    }
+
+    fun messages(): List<DecodedMessage> {
+        return runBlocking {
+            libXMTPGroup.sync()
```
nplasterer, line 84: I'm assuming we probably don't want to put the sync in the message listing? And leave it up to the client to remember to sync before displaying messages?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -81,6 +89,27 @@ data class Conversations(
...
+        return Group(client, group)
+    }
+
+    suspend fun syncGroups() {
+        libXMTPConversations?.sync()
+    }
+
+    fun listGroups(): List<Group> {
+        return runBlocking {
+            syncGroups()
```
nplasterer, line 106: Same here. Probably don't want to run sync for them?

library/src/androidTest/java/org/xmtp/android/library/GroupMembershipChangeTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        val messages = group.messages()
+        assertEquals(messages.size, 1)
+        assertEquals(group.memberAddresses().size, 3)
+        group.removeMembers(listOf(caro.walletAddress.lowercase()))
+        assertEquals(group.messages().size, 2)
+        assertEquals(group.memberAddresses().size, 2)
+        val content: GroupMembershipChanges? = messages.first().content()
+        assertEquals(
+            listOf(caro.walletAddress.lowercase()),
+            content?.membersRemovedList?.map { it.accountAddress.lowercase() }?.sorted()
```
nplasterer, line 93: This is failing saying `membersRemovedList` is empty. 🤔

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -50,16 +56,28 @@ sealed class Conversation {
             return when (this) {
                 is V1 -> conversationV1.peerAddress
                 is V2 -> conversationV2.peerAddress
+                is Group -> group.memberAddresses().joinToString(",")
```
nplasterer, line 59: I'm returning a comma separated list here but I could also throw an error message to use `peerAddresses` instead.

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -68,15 +86,16 @@ sealed class Conversation {
...
         }
 
     fun consentState(): ConsentState {
-        val client: Client = when (this) {
-            is V1 -> conversationV1.client
-            is V2 -> conversationV2.client
+        return when (this) {
+            is V1 -> conversationV1.client.contacts.consentList.state(address = peerAddress)
+            is V2 -> conversationV2.client.contacts.consentList.state(address = peerAddress)
+            is Group -> ConsentState.UNKNOWN // No such thing as consent for a group
```
nplasterer, line 97: Or should this always have the consent state as Allowed for now?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -32,11 +36,15 @@ import org.xmtp.proto.keystore.api.v1.Keystore.TopicMap.TopicData
...
 import org.xmtp.android.library.messages.DecryptedMessage
+import uniffi.xmtpv3.FfiConversations
+import uniffi.xmtpv3.FfiListConversationsOptions
+import uniffi.xmtpv3.org.xmtp.android.library.libxmtp.GroupEmitter
 import java.util.Date
 
 data class Conversations(
     var client: Client,
     var conversationsByTopic: MutableMap<String, Conversation> = mutableMapOf(),
+    private val libXMTPConversations: FfiConversations? = null,
```
nplasterer, line 47: Technically I could get this from the client via `client.v3Client.conversations` But I kind of liked passing it explicitly. I could go either way.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -0,0 +1,204 @@
...
+    fun testCanStreamGroupsAndConversations() = kotlinx.coroutines.test.runTest {
+        boClient.conversations.stream(includeGroups = true).test {
+            val group =
+                boClient.conversations.newGroup(listOf(alix.walletAddress.lowercase()))
+            val conversation =
+                boClient.conversations.newConversation(alix.walletAddress.lowercase())
+            assertEquals("hi", awaitItem().messages().first().body)
+            awaitComplete()
+        }
+    }
```
nplasterer, line 203: Non of my streaming tests pass... so I'm either doing the tests wrong or I'm setting up streaming wrong.

library/src/androidTest/java/org/xmtp/android/library/GroupMembershipChangeTest.kt
```diff
@@ -0,0 +1,96 @@
...
+                caro.walletAddress.lowercase()
+            )
+        )
+        val messages = group.messages()
+        assertEquals(messages.size, 1)
+        assertEquals(group.memberAddresses().size, 3)
+        group.removeMembers(listOf(caro.walletAddress.lowercase()))
+        assertEquals(group.messages().size, 2)
+        assertEquals(group.memberAddresses().size, 2)
+        val content: GroupMembershipChanges? = messages.first().content()
```
neekolas, line 89: Shouldn't we be looking for the second item in the array here, not the first?

library/src/main/java/org/xmtp/android/library/Conversation.kt
neekolas: I'm OK with throwing an error here. It should only come up when you have enabled the groups flag. Definitely worth calling out in the release notes if we go that way, though.

library/src/main/java/org/xmtp/android/library/Conversations.kt
neekolas: It might depend on whether we want to allow people to page through the `listGroups` results. If you're paging, we definitely don't want to re-check on each page. If we're always returning the entire list (setting null on the options below), maybe it's fine to do the check implicitly?

library/src/main/java/org/xmtp/android/library/libxmtp/GroupEmitter.kt
```diff
@@ -0,0 +1,17 @@
...
+import uniffi.xmtpv3.FfiConversationCallback
+import uniffi.xmtpv3.FfiGroup
+
+class GroupEmitter {
+    private val _groups = MutableSharedFlow<FfiGroup>()
+    val groups = _groups.asSharedFlow()
+
+    val callback: FfiConversationCallback = object : FfiConversationCallback {
+        override fun onConversation(conversation: FfiGroup) {
+            _groups.tryEmit(conversation)
```
neekolas, line 14: If you're having trouble with streaming, might be worth logging something here so we can isolate if the problem is in Rust or Kotlin. Totally possible something is locking up in Rust

library/src/main/java/org/xmtp/android/library/Group.kt
neekolas: That makes sense to me. I would expect people to use limit/before/after to page through results here. These lists could be very long.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -0,0 +1,208 @@
...
+
+    @Test
+    fun testCanStreamGroups() = kotlinx.coroutines.test.runTest {
+        boClient.conversations.streamGroups().test {
+            val conversation =
+                boClient.conversations.newGroup(listOf(alix.walletAddress.lowercase()))
+            conversation.send(content = "hi")
+            assertEquals("hi", awaitItem().messages().first().body)
```
neekolas, line 192: I'd expect "hi" to be the second item, after the membership change message

nplasterer: The messaging list is in descending order by default. So what ever the last message to be sent is the first message.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -334,10 +337,12 @@ class Client() {
 
         if (v3Client?.textToSign() == null) {
             v3Client?.registerIdentity(null)
-        } else {
+        } else if (account != null) {
             v3Client.textToSign()?.let {
                 v3Client.registerIdentity(account.sign(it))
             }
+        } else {
+            Log.i(TAG, "No signer passed but signer was required.")
```
nplasterer, line 345: Just soft error for now instead of blowing up if they try to create a bundle without a signer and don't already have v3 keys.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
neekolas: Hmmm. Then I don't know. Are you able to see any messages come into Kotlin-land?

nplasterer: The streaming doesn't seem to be working. But I can list messages no problem and sync. I think this is just a streaming issue.

library/src/androidTest/java/org/xmtp/android/library/GroupMembershipChangeTest.kt
nplasterer: Opened an issue in LibXMTP for this https://github.com/xmtp/libxmtp/issues/464

library/src/main/java/org/xmtp/android/library/Conversation.kt
nplasterer: I think I'm just going to leave this as a string for now. Less breaking changes for those not using groups and then when we roll it out I can do more things that break it.

library/src/androidTest/java/org/xmtp/android/library/GroupMembershipChangeTest.kt
nplasterer: No because the list is reversed because the messages come in descending order.

library/src/main/java/org/xmtp/android/library/Conversation.kt
richardhuaaa: Assuming that implementors are not going to persist it, it makes sense to hardcode it to ALLOWED, so that it shows up correctly in their UI, and then use the true value once we have implemented consent

nplasterer: I actually think its probably safer to keep this unknown for now since when they move from request to inbox it would feel like a broken experience once it changes to the network based version. Probably should prioritize group chat consent sooner. 🤔

neekolas: Group chat consent doesn't seem that bad. I'm tempted to sneak it in before Denver (on top of our existing consent mechanism, just with some new values)

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -99,13 +124,16 @@ sealed class Conversation {
                             .setKeyMaterial(conversationV2.keyMaterial.toByteString()),
                     ),
             ).build()
+
+            is Group -> throw XMTPException("Groups do not support topics")
```
richardhuaaa, line 128: Why is `TopicData` and `keyMaterial` exposed in v2? What will integrators do with it?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -81,6 +86,45 @@ data class Conversations(
...
+    fun newGroup(accountAddresses: List<String>): Group {
+        if (accountAddresses.isEmpty()) {
+            throw XMTPException("Cannot start an empty group chat.")
+        }
+        if (accountAddresses.size == 1 &&
+            accountAddresses.first().lowercase() == client.address.lowercase()
+        ) {
+            throw XMTPException("Recipient is sender")
+        }
+        val contacts = accountAddresses.map { client.contacts.find(it) }
```
richardhuaaa, line 98: Is this only checking for the existence of a v2 bundle?

Not a blocker here, but we're going to have to update this here to check if the recipient has a v3 installation on the network. We probably need to expose this from libxmtp if we haven't already

nplasterer: Good call. I think there is a canMessage function already I'll expose that and us that here instead.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -552,6 +559,13 @@ class Client() {
         return runBlocking { query(Topic.contact(peerAddress)).envelopesList.size > 0 }
     }
 
+    fun canMessage(addresses: List<String>): Boolean {
+        return runBlocking {
+            libXMTPClient != null && !libXMTPClient!!.canMessage(addresses.map { it.lowercase() })
```
nplasterer, line 564: @richardhuaaa my tests failed after I implemented this. I think this is another scenario where lowercase should be handled on the rust side.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -324,7 +327,7 @@ class Client() {
                     isSecure = false,
                     db = dbPath,
                     encryptionKey = retrievedKey.encoded,
-                    accountAddress = account.address.lowercase(),
+                    accountAddress = accountAddress.lowercase(),
```
nplasterer, line 330: I should remove the lowercase here as well.

richardhuaaa: No changes needed in libxmtp for this one right?

nplasterer: @richardhuaaa just removed this lowercase and now the tests fail so I think something on the rust side could be helpful. But not a huge deal we can leave the lowercaseing here

library/src/main/java/org/xmtp/android/library/Conversation.kt
nplasterer: I believe `TopicData` and `keyMaterial` is used for push notifications. That way they can export from react native or javascript and import into iOS or android and not need to reload the entire costly `conversation.list()` method to decrypt a push notification.

library/src/main/java/org/xmtp/android/library/Client.kt
richardhuaaa: What was the error message?

richardhuaaa: Never mind, found it

library/src/main/java/org/xmtp/android/library/Client.kt
richardhuaaa: https://github.com/xmtp/libxmtp/pull/468

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -127,6 +155,8 @@ sealed class Conversation {
             is V2 -> {
                 conversationV2.prepareMessage(content = content, options = options)
             }
+
+            is Group -> throw XMTPException("Groups do not support prepared messages") // We return a encoded content not a preparedmessage which requires a envelope
```
richardhuaaa, line 159: I'm not a kotlin expert and am not familiar with sealed classes. Are these methods being exposed directly to implementors, and are they required to know which methods they can/can't call on groups?

If so, can we find a way to use strong typing (for example, separate types for group and 1:1) where only the methods they can call are exposed?

richardhuaaa (approved):

library/src/main/java/org/xmtp/android/library/Conversation.kt
nplasterer: Ya I definitely have thought about this and maybe it can be spike/refactor. Since groups aren't being created users shouldn't experience this for now. But agree it almost feels like we need to return a type of either Conversation or Group instead of making Groups conform to conversations but it would be a little extra work. My hope is when we get rid of V2 this won't be a problem because most all the same functions should exist for both groups and 1to1 in that world.

PR #161, opened 2024-01-24
fix: add catch to listBatchDecryptedMessages method

kele-leanes:
Closes https://github.com/xmtp/xmtp-react-native/issues/218.

This update adds a try-catch block to the `listBatchDecryptedMessages` method, improving error handling.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -358,8 +358,13 @@ data class Conversations(
                                 Log.d(TAG, "discarding message, unknown conversation $envelope")
                                 return@mapNotNull null
                             }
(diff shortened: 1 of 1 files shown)


nplasterer (approved):

nplasterer:
I bet this change should also be made to `listBatchMessages` as well.

kele-leanes:
I wonder if this change is necessary on `listBatchMessages` since [decodeOrNull](https://github.com/xmtp/xmtp-android/blob/1a1e055f6af45880720890ff3d0fe2edebef69a2/library/src/main/java/org/xmtp/android/library/Conversation.kt#L112) is already catching the possible error

nplasterer:
Ahh nice catch I'll roll back my addition!

PR #164, opened 2024-01-26
Add sender HMAC to MessageV2

giovasdistillery:
added shouldPush to ContentCodec type
added shouldPush method to TextCodec and CompositeCodec
added senderHmac to MessageV2
added encryption methods for calculating HMAC signature


library/src/main/java/org/xmtp/android/library/messages/MessageV2.kt
```diff
@@ -18,7 +19,12 @@ typealias MessageV2 = org.xmtp.proto.message.contents.MessageOuterClass.MessageV
 
 class MessageV2Builder {
     companion object {
-        fun buildFromCipherText(headerBytes: ByteArray, ciphertext: CipherText?): MessageV2 {
+        fun buildFromCipherText(
+            headerBytes: ByteArray,
+            ciphertext: CipherText?,
+            senderHmac: ByteArray?,
+            shouldPush: Boolean,
```
nplasterer, line 28: So unfortunately android doesn't make it as easy to extend MessageV2 with new fields like they can in iOS and Swift. 
I'm going to push up a commit with what I think my be necessary.

nplasterer (commented):
I pushed up a commit to hopefully unblock you. I think adding a test to make sure we can get the shouldPush method like we can with fallbacks would be much appreciated. 🙏

library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -77,26 +77,35 @@ class Crypto {
...
+        }
+
+        fun generateHmacSignature(
+            secret: ByteArray,
+            info: ByteArray,
+            message: ByteArray,
+        ): ByteArray {
+            val hkdfKey = deriveKey(secret, message, info)
+            return calculateMac(secret, hkdfKey)
+        }
```
nplasterer, line 109: @Bren2010 do you mind giving this code a look to make sure it's correct. 🙏

nplasterer (commented):
This is looking really good. Just have some concerns about the cryptography. Also Rich mentioned that this might be missing the method `getV2ConversationHmacKeys()` https://github.com/xmtp/xmtp-js/pull/519/files#diff-3f8d138711ab8a38267eef7baab8b362cbe3deee7fda5e18689bae9c2db74bb9

library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -77,26 +77,35 @@ class Crypto {
...
+            return hkdf
+        }
+
+        fun generateHmacSignature(
+            secret: ByteArray,
+            info: ByteArray,
+            message: ByteArray,
+        ): ByteArray {
+            val hkdfKey = deriveKey(secret, message, info)
+            return calculateMac(secret, hkdfKey)
```
Bren2010, line 108:
```suggestion
            val hkdfKey = deriveKey(secret, info)
            return calculateMac(hkdfKey, message)
```

library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -77,26 +77,35 @@ class Crypto {
...
-        val derivationParameters = HKDFParameters(secret, salt, info)
-        val digest = SHA256Digest()
-        val hkdfGenerator = HKDFBytesGenerator(digest)
-        hkdfGenerator.init(derivationParameters)
-        val hkdf = ByteArray(32)
-        hkdfGenerator.generateBytes(hkdf, 0, hkdf.size)
-        return hkdf
+        fun deriveKey(
+            secret: ByteArray,
+            salt: ByteArray,
```
Bren2010, line 90:
```suggestion
```

library/src/main/java/org/xmtp/android/library/Crypto.kt
```diff
@@ -77,26 +77,35 @@ class Crypto {
...
+            info: ByteArray,
+        ): ByteArray {
+            val derivationParameters = HKDFParameters(secret, salt, info)
```
Bren2010, line 93:
```suggestion
            val derivationParameters = HKDFParameters(secret, ByteArray(0), info)
```

Bren2010 (changes requested):

nplasterer:
Merging in @kele-leanes https://github.com/xmtp/xmtp-android/pull/184 to unblock RN looks to have broken a few places in the code. We can revert this PR if there is a better approach to how to support RN.

library/src/androidTest/java/org/xmtp/android/library/CodecTest.kt
```diff
@@ -109,4 +119,97 @@ class CodecTest {
...
+            client = aliceClient,
+            encodedContent = messages[0].encodedContent,
+            topic = aliceConversation.topic,
+            keyMaterial = aliceConversation.keyMaterial!!,
+            codec = codec,
+        )
+
+        assertEquals(false, message.shouldPush)
+        assertEquals(true, message.senderHmac?.isNotEmpty())
+        val keys = aliceClient.conversations.getHmacKeys()
```
rygine, line 148: can this be removed? doesn't look like it's being used.

library/src/androidTest/java/org/xmtp/android/library/CodecTest.kt
```diff
@@ -109,4 +119,97 @@ class CodecTest {
...
+
+        val topics = hmacKeys.hmacKeysMap.keys
+        conversations.forEach { convo ->
+            assertTrue(topics.contains(convo.topic))
+        }
+
+        val topicHmacs = mutableMapOf<String, ByteArray>()
+        val headerBytes = ByteArray(10)
+
+        conversations.map { conversation ->
```
rygine, line 182: nit: since this isn't returning anything, is `forEach` more appropriate here?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -289,6 +294,45 @@ data class Conversations(
         return conversation
     }
 
+    fun getHmacKeys(
+        request: Keystore.GetConversationHmacKeysRequest? = null,
+    ): Keystore.GetConversationHmacKeysResponse {
+        val thirtyDayPeriodsSinceEpoch = (Date().time / 1000 / 60 / 60 / 24 / 30).toInt()
```
rygine, line 300: is there a reason this is different than the previously used value?

```
Instant.now().epochSecond / 60 / 60 / 24 / 30
```

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -289,6 +294,45 @@ data class Conversations(
...
+        if (!request?.topicsList.isNullOrEmpty()) {
+            topics = topics.filter {
+                request!!.topicsList.contains(it.key)
+            }.toMutableMap()
+        }
+
+        topics.forEach {
+            val conversation = it.value
+            if (conversation.keyMaterial != null) {
+                (thirtyDayPeriodsSinceEpoch - 1..thirtyDayPeriodsSinceEpoch + 1).map { value ->
```
rygine, line 314: why `map` here isn't of `forEach`?

library/src/main/java/org/xmtp/android/library/SendOptions.kt
```diff
@@ -5,5 +5,6 @@ import org.xmtp.proto.message.contents.Content
 data class SendOptions(
     var compression: EncodedContentCompression? = null,
     var contentType: Content.ContentTypeId? = null,
-    var ephemeral: Boolean = false
+    var ephemeral: Boolean = false,
+    var __shouldPush: Boolean? = null
```
rygine, line 9: why is `shouldPush` a send option? i thought the codecs are handling this value.

library/src/main/java/org/xmtp/android/library/messages/DecryptedMessage.kt
```diff
@@ -8,5 +8,6 @@ data class DecryptedMessage(
     var encodedContent: EncodedContent,
     var senderAddress: String,
     var sentAt: Date,
-    var topic: String = ""
+    var topic: String = "",
+    var shouldPush: Boolean? = null,
```
rygine, line 12: why wouldn't this always be a boolean with a default value of `false`?

library/src/main/java/org/xmtp/android/library/SendOptions.kt
nplasterer: @kele-leanes I agree on this we should be able to get if it shouldPush from the contentType which is part of the send options

kele-leanes: hi @rygine here is the [discussion](https://github.com/xmtp/xmtp-react-native/pull/235#discussion_r1484722744) where we came up with this approach

library/src/main/java/org/xmtp/android/library/Conversations.kt
nplasterer: I was just copying how you did it here: https://github.com/xmtp/xmtp-js/pull/519/files#diff-3f8d138711ab8a38267eef7baab8b362cbe3deee7fda5e18689bae9c2db74bb9R306

library/src/main/java/org/xmtp/android/library/messages/DecryptedMessage.kt
kele-leanes: we are checking it here https://github.com/xmtp/xmtp-android/pull/164/commits/887f1f4dce979c57cf4d86d2fd035273fd70d78d#diff-5beaf5d0240fb991d7984f4f1dba63dcbafa3138e84bd5d510a688ad83446806R164:~:text=val-,calculatedShouldPush,-%3D%20shouldPush%20%3F

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -417,6 +429,7 @@ class GroupTest {
     }
 
     @Test
+    @Ignore("CI Issues")
     fun testCanStreamGroupsAndConversations() = kotlinx.coroutines.test.runTest {
         boClient.conversations.streamAll().test {
             val group =
```
nplasterer, line 435: All the changes in this file need to be reverted.

PR #166, opened 2024-01-30
Group Chat Streaming

nplasterer:
This adds the streaming components of group chat
- stream messages
- stream groups
- stream both groups and conversations together


library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -136,4 +141,26 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
...
+        }
+
+        val stream = libXMTPGroup.stream(messageCallback)
+        awaitClose { stream.end() }
+    }
+
+    fun streamDecryptedMessages(): Flow<DecryptedMessage> = callbackFlow {
+        val messageCallback = object : FfiMessageCallback {
+            override fun onMessage(message: FfiMessage) {
+                trySend(decrypt(Message(client, message)))
```
neekolas, line 159: I think we want to refer to what you're doing in here as decoding rather than decrypting. The messages all get decrypted in `libxmtp`.

nplasterer: So we have two streaming methods one for decoding and one for decrypting. We need this stream decrypted messages for React Native custom content types. @nakajima could speak more to it.

Since Groups are conforming to the current Conversation model I have to use these methods. Once we do the refactor to Groups Dms and Conversation parent we might be able to get rid of some of these things.

richardhuaaa: What is the intended difference between a `DecryptedMessage` and a `DecodedMessage`? When I trace the method calls from both streaming methods, the `DecryptedMessage` one ultimately gives `message.decode().encodedContent`, and the `DecodedMessage` one ultimately gives `EncodedContent.parseFrom(message.content)`

richardhuaaa (approved):
Have a question in the PR but accepting to unblock React Native, well done on the persistence and fast turnaround! Did you manage to get both streams working simultaneously?

library/src/main/java/org/xmtp/android/library/Group.kt
nplasterer: I don't think there is going to be any difference in the V3 world but again I didn't follow as closely to the reasoning for custom contentTypes. I found this for context https://github.com/xmtp/xmtp-ios/pull/196

PR #170, opened 2024-02-05
Allow LOCAL with a custom string.

nplasterer:
You can call this now with `XMTPEnvironment.LOCAL.withValue("<IP ADDRESS>")`

Gives the ability to run a physical android app pointing at a local environment.


cameronvoell (approved):
looks good!

PR #171, opened 2024-02-05
Remove legacy sign method

nplasterer:
This removes the `legacySign` method introduced in a previous PR to just get the bytes of the signature out of the previous sign method to limit breaking changes.


cameronvoell (approved):
verified this allows react-native main branch to build immediately with the branch here. looks good!

PR #174, opened 2024-02-09
Stream Groups Returns Groups

nplasterer:
Adds a new method that returns groups and makes the stream conversation private.


cameronvoell (approved):
looks good 👍

PR #176, opened 2024-02-13
Add isActive to Groups

nplasterer:
This adds a check to see if the group is active or if the member has been removed so that we can update the UI accordingly.

This also adds libxmtp versioning to help with debugging and fixes a client creation issue.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -349,7 +349,7 @@ class Client() {
                 throw XMTPException("No signer passed but signer was required.")
             }
         }
-        Log.i(TAG, "The libxmtp version is: $libXMTPVersion")
+        Log.i(TAG, "LibXMTP $libXMTPVersion")
         return v3Client
     }
 
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -44,6 +44,7 @@ import org.xmtp.proto.message.api.v1.MessageApiOuterClass.QueryRequest
 import uniffi.xmtpv3.FfiXmtpClient
 import uniffi.xmtpv3.LegacyIdentitySource
 import uniffi.xmtpv3.createClient
+import uniffi.xmtpv3.getVersionInfo
 import java.io.File
 import java.nio.charset.StandardCharsets
 import java.security.KeyStore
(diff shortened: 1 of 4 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -81,6 +82,7 @@ class Client() {
     lateinit var conversations: Conversations
     var logger: XMTPLogger = XMTPLogger()
     var libXMTPClient: FfiXmtpClient? = null
+    val libXMTPVersion: String = getVersionInfo()
```
nplasterer, line 85: Do you think it would be helpful to have this field on here so we can get it on Client directly? Or just the log? @richardhuaaa

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -128,6 +128,11 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         )
     }
 
+    fun isActive(): Boolean {
+        runBlocking { libXMTPGroup.sync() }
```
nplasterer, line 132: I don't think there should be any reason you'd want a cached version of isActive. So syncing before hand.

library/src/main/java/org/xmtp/android/library/Client.kt
richardhuaaa: Don't have a strong opinion, that works!

library/src/main/java/org/xmtp/android/library/Group.kt
neekolas: Depends on how people are using this. If you are running through a long list of conversations and calling is_active on all of them regularly, that can be pretty expensive.

nplasterer: Interesting okay 🤔 Ya I could see that. So really only sync if they click on a group to make sure they can see it.

neekolas (approved):

richardhuaaa (approved):

library/src/main/java/xmtpv3.kt
```diff
@@ -413,6 +413,8 @@ internal interface _UniFFILib : Library {
     ): RustBuffer.ByValue
```
richardhuaaa, line 413: There should have been a `libxmtp-version.txt` generated alongside the `.kt` definition, could we check that in too?

PR #180, opened 2024-02-16
group preference actions

tuddman:
## Summary

With https://github.com/xmtp/proto/pull/146 in, this patch adds initial support for recognizing group consents:   `AllowGroup` and `DenyGroup`.

Bumps `org.xmtp:proto-kotlin` to `3.43.2`

Fixes some other kotlin warnings


library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -45,70 +54,103 @@ class ConsentList(val client: Client) {
...
-                    payload.toUByteArray().toByteArray()
-                )
+                    payload.toUByteArray().toByteArray(),
+                ),
             )
         }
 
         preferences.iterator().forEach { preference ->
-            preference.allow?.walletAddressesList?.forEach { address ->
+            preference.allowDm?.walletAddressesList?.forEach { address ->
```
nplasterer, line 87: I probably should have commented this on the proto PR. But you're denying an address in this case so it should honestly be `allowAddress` or `denyAddress` since we also use this functionality inside groups to change the way messages appear in a group context for a denied address. But this doesn't get surfaced to the user so probably not a huge deal.

nplasterer (commented):
This looks really great! Thanks for doing this. A couple of notes
Can you update this location to correctly reflect the groups consent state https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/Conversation.kt#L103

Can you add in the implicit consent like we do for [DMs](https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/Conversations.kt#L215) on `newGroup` https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/Conversations.kt#L94-L116

We also do implicit consent when [sending a message](https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/ConversationV2.kt#L194) to a conversation with a state of unknown so we should probably also do that for groups. https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/Group.kt#L40

library/src/main/java/org/xmtp/android/library/Contacts.kt
tuddman: Please see https://github.com/xmtp/proto/pull/149 and consider ✅  that one first.  And I'll use update this PR to leverage that one.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -342,4 +342,41 @@ class GroupTest {
...
+        val group = boClient.conversations.newGroup(
+            listOf(
+                alix.walletAddress,
+                caro.walletAddress
+            )
+        )
+
+        var result = boClient.contacts.isGroupAllowed(group.id)
+
+        assert(!result)
```
nplasterer, line 357: I assumed that this test should fail now since the default is allow on newGroup?

nplasterer (approved):
The code looks great. Approving to unblock.
Would like to see some additional testing around the implicit consent stuff
- a check here that the group has a state of allowed https://github.com/xmtp/xmtp-android/blob/main/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt#L267-L280
- a check here that the bogroup has a state of allowed and alixgroup has a state of unknown https://github.com/xmtp/xmtp-android/blob/main/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt#L92-L132

PR #183, opened 2024-02-18
Improve V3 Client Create

nplasterer:
This allows the passing of your own database URL and encryption keys if so desired.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -578,11 +578,12 @@ class Client() {
         return runBlocking { query(Topic.contact(peerAddress)).envelopesList.size > 0 }
     }
 
-    fun canMessage(addresses: List<String>): Boolean {
-        return runBlocking {
-            libXMTPClient != null && !libXMTPClient!!.canMessage(addresses.map { it })
-                .contains(false)
+    fun canMessageV3(addresses: List<String>): Boolean {
+        if (libXMTPClient == null) return false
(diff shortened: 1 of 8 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -81,8 +83,8 @@ class Client() {
     lateinit var contacts: Contacts
     lateinit var conversations: Conversations
     var logger: XMTPLogger = XMTPLogger()
-    var libXMTPClient: FfiXmtpClient? = null
     val libXMTPVersion: String = getVersionInfo()
+    private var libXMTPClient: FfiXmtpClient? = null
```
nplasterer, line 87: I think this should be private. Clients shouldn't be able to access the V3 Rust stuff directly by passing any sdk level checks we have on them.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -564,11 +578,12 @@ class Client() {
         return runBlocking { query(Topic.contact(peerAddress)).envelopesList.size > 0 }
     }
 
-    fun canMessage(addresses: List<String>): Boolean {
-        return runBlocking {
-            libXMTPClient != null && !libXMTPClient!!.canMessage(addresses.map { it })
-                .contains(false)
+    fun canMessageV3(addresses: List<String>): Boolean {
```
nplasterer, line 581: Make this more inline with the naming in iOS

neekolas (approved):

PR #187, opened 2024-02-22
feat: add function to delete database

nplasterer:
Method to delete database that can be called on wallet disconnection.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -82,6 +82,7 @@ class Client() {
     lateinit var apiClient: ApiClient
     lateinit var contacts: Contacts
     lateinit var conversations: Conversations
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -160,6 +161,7 @@ class Client() {
         privateKeyBundleV1: PrivateKeyBundleV1,
         apiClient: ApiClient,
         libXMTPClient: FfiXmtpClient? = null,
+        dbPath: String = ""
```
nplasterer, line 164: Instead of having to store the dbPath like this is it something the `FfiXmtpClient` could expose?

neekolas: We could keep track of that and expose it. We have the info

richardhuaaa (approved):

PR #191, opened 2024-02-29
feat: streaming and threading improvements

nplasterer:
This does a major overhaul of how we are doing async calls in Android. Switching from `runBlocking` to `suspend` functions for all send message functionality to help prevent deadlocking during streaming.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversation.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversation.kt
@@ -169,7 +169,7 @@ sealed class Conversation {
         }
     }
 
-    fun send(prepared: PreparedMessage): String {
+    suspend fun send(prepared: PreparedMessage): String {
         return when (this) {
             is V1 -> conversationV1.send(prepared = prepared)
             is V2 -> conversationV2.send(prepared = prepared)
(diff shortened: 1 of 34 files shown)


library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -169,31 +169,31 @@ sealed class Conversation {
         }
     }
 
-    fun send(prepared: PreparedMessage): String {
+    suspend fun send(prepared: PreparedMessage): String {
```
richardhuaaa, line 172: Is this a breaking change, or will existing code continue to work (albeit no longer waiting for the publish to complete)? Not in this PR, but long-term do we have any thoughts on splitting out the two conversation types, and allowing the API's to be different across the two to maintain backwards compatibility?

richardhuaaa (approved):

library/src/main/java/org/xmtp/android/library/Conversation.kt
nplasterer: Technically it's a breaking change if anyone was using send outside of a thread. As you can see there are no changes to the example app in this PR because my example app was already calling send in a thread so I honestly think this change may be unnoticeable to most integrators as they were likely sending it asynchronously anyways.

PR #195, opened 2024-03-06
thirtyDayPeriodsSinceEpoch update

nplasterer:
In JS they do this

```js
const thirtyDayPeriodsSinceEpoch = Math.floor(
          Date.now() / 1000 / 60 / 60 / 24 / 30
        )
```

I think this brings the code closer inline with that.


kele-leanes (approved):

nplasterer:
We validated that this already works so going to close this.

PR #210, opened 2024-03-28
Group Chat Push Notifications

nplasterer:
Adds the ability to get push notifications from group chats and group chat welcomes

https://github.com/xmtp/xmtp-android/assets/3522670/284f75d3-78f2-4d91-b8b0-2d87dab094f3


example/src/main/java/org/xmtp/android/example/MainViewModel.kt
```diff
@@ -56,10 +57,20 @@ class MainViewModel : ViewModel() {
                     }
 
                     Service.Subscription.newBuilder().also { sub ->
-                        sub.addAllHmacKeys(result)
+                        if (!result.isNullOrEmpty()) {
+                            sub.addAllHmacKeys(result)
+                        }
```
nplasterer, line 62: The hmac key work for groups will come in a follow up.

cameronvoell (approved):

PR #213, opened 2024-04-02
feat: Frames Client

alexrisch:
Added Frames package port from JS
Bumped protos library


library/src/main/java/org/xmtp/android/library/frames/FramesTypes.kt
```diff
@@ -0,0 +1,103 @@
...
+    val acceptedClients: AcceptedFrameClients,
+    val image: OpenFrameImage,
+    val postUrl: String?,
+    val textInput: TextInput?,
+    val buttons: Map<String, OpenFrameButton>?,
+    val ogImage: String,
+    val state: String?
+)
+
+data class GetMetadataResponse(
```
neekolas, line 41: It's going to make usability much better if this also included the `frameInfo` field which does a lot of the parsing of the extracted tags for you and assembles it into a nice object. Could happen in a follow-up PR.

neekolas (approved):
This looks great

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.xmtp.android.library.messages.Signature
+import org.xmtp.android.library.messages.getPublicKeyBundle
+import org.xmtp.proto.message.contents.PublicKeyOuterClass.SignedPublicKeyBundle
+import java.security.MessageDigest
+import org.xmtp.proto.message.contents.Frames.FrameActionBody
+import org.xmtp.proto.message.contents.Frames.FrameAction
+
+class FramesClient(private val xmtpClient: Client, var proxy: OpenFramesProxy = OpenFramesProxy()) {
+
+    suspend fun signFrameAction(inputs: FrameActionInputs): FramePostPayload {
```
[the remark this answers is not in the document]
cameronvoell: I don't believe you need withContext here. Whoever calls the suspend function could launch a coroutine and set the thread using withContext (or just using runBlocking is suitable in test scenarios like in your test).

cameronvoell (approved):

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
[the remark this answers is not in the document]
nplasterer: Ya what @cameronvoell said. We want to leave it up to the developer to decide if they want to run this in a background or the main thread. So leaving it out makes sense.

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,92 @@
...
+            .setUnixTimestamp(now.toInt())
+
+        if (inputText != null) {
+            frameActionBuilder.inputText = inputText
+        }
+        if (state != null) {
+            frameActionBuilder.state = state
+        }
+
+        val toSign = frameActionBuilder.build()
```
nplasterer, line 38: This is a total fine way of doing this but in all the other places in the code we actually do it like this

```suggestion
        val toSign = FrameActionBody.newBuilder().also { frame ->
            frame.frameUrl = frameUrl
            frame.buttonIndex = buttonIndex
            frame.opaqueConversationIdentifier = opaqueConversationIdentifier
            frame.timestamp = now
            frame.unixTimestamp = now.toInt()
            frame.inputText  = inputText
            frame.state = state
        }.build()
```

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    private suspend fun buildSignedFrameAction(actionBodyInputs: FrameActionBody): ByteArray {
+        val digest = sha256(actionBodyInputs.toByteArray())
+        val signature = signDigest(digest)
+
+        val publicKeyBundle = getPublicKeyBundle()
+        val frameAction = FrameAction.newBuilder()
+            .setActionBody(actionBodyInputs.toByteString())
+            .setSignature(signature)
+            .setSignedPublicKeyBundle(publicKeyBundle).build()
```
nplasterer, line 65: Same here. Setters are a very Java way of doing things. Kotlin prefers this method

```suggestion
        val frameAction = FrameAction.newBuilder().also {
            it.actionBody = actionBodyInputs.toByteString()
            it.signature = signature
            it.signedPublicKeyBundle = publicKeyBundle
         }.build()
```

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,92 @@
...
+        return when (inputs.conversationInputs) {
+            is ConversationActionInputs.Group -> {
+                val groupInputs = inputs.conversationInputs.inputs
+                val combined = groupInputs.groupId + groupInputs.groupSecret
+                val digest = sha256(combined)
+                Base64.encodeToString(digest, Base64.NO_WRAP)
+            }
+            is ConversationActionInputs.Dm -> {
+                val dmInputs = inputs.conversationInputs.inputs
+                val conversationTopic = dmInputs.conversationTopic ?: throw InvalidArgumentsError()
```
nplasterer, line 80: Can we throw an XMTPException instead. This helps integrators catch exceptions being thrown from our SDK easier.
```suggestion
                val conversationTopic = dmInputs.conversationTopic ?: throw XMTPException("No conversation topic")
```

library/src/main/java/org/xmtp/android/library/frames/FramesErrors.kt
```diff
@@ -0,0 +1,5 @@
+package org.xmtp.android.library.frames
+
+class FramesApiError(message: String, val status: Int) : Exception(message)
```
nplasterer, line 3: I think you can just delete this whole file. But if you want to keep it then I would have it extend the `XMTPException` instead of exception.

library/src/main/java/org/xmtp/android/library/frames/OpenFramesProxy.kt
```diff
@@ -0,0 +1,26 @@
...
+    suspend fun post(url: String, payload: FramePostPayload): GetMetadataResponse {
+        return inner.post(url, payload)
+    }
+
+    suspend fun postRedirect(url: String, payload: FramePostPayload): FramesApiRedirectResponse {
+        return inner.postRedirect(url, payload)
+    }
+
+    fun mediaUrl(url: String): String {
+        if (url.startsWith("data:")) {
```
nplasterer, line 20: I think this is fine but sometimes it can get a little funky. I like to lean on the convenience methods like 

`URI(url).scheme == "data"`

library/src/androidTest/java/org/xmtp/android/library/FramesTest.kt
```diff
@@ -0,0 +1,71 @@
...
+class FramesTest {
+    @Test
+    fun testFramesClient() {
+        val frameUrl = "https://fc-polls-five.vercel.app/polls/01032f47-e976-42ee-9e3d-3aac1324f4b8"
+        val fixtures = fixtures()
+        val aliceClient = fixtures.aliceClient
+
+        val framesClient = FramesClient(xmtpClient = aliceClient)
+        val conversationTopic = "foo"
+        val participantAccountAddresses = listOf("amal", "bola")
```
nplasterer, line 28: In case you haven't seen we have a new cast https://github.com/jhaaaa/alix-and-bo

nplasterer (commented):
This looks really great. Mostly all nits to keep it inline with the rest of the kotlin sdk.

Really don't know how to review the `ProxyClient`. Kind of feels like something that might have been better suited for GRPC. But I might be missing it's use case.

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,94 @@
...
+            frame.buttonIndex = buttonIndex
+            frame.opaqueConversationIdentifier = opaqueConversationIdentifier
+            frame.timestamp = now
+            frame.unixTimestamp = now.toInt()
+            if (inputText != null) {
+                frame.inputText = inputText
+            }
+            if (state != null) {
+                frame.state = state
+            }
```
nplasterer, line 36: Is there a reason you null check here? Seems like state and inputText are okay with null being set?

```suggestion
                frame.inputText = inputText
                frame.state = state
```

nplasterer (approved):

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
alexrisch: They are not, it errors if the state is null or inputText is null, instead they use the default values

nplasterer: Ahh okay now I see you did this in iOS

`let state = inputs.state ?? ""`
 Then to clean this up what about this

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Date
+
+class FramesClient(private val xmtpClient: Client, var proxy: OpenFramesProxy = OpenFramesProxy()) {
+
+    suspend fun signFrameAction(inputs: FrameActionInputs): FramePostPayload {
+        val opaqueConversationIdentifier = buildOpaqueIdentifier(inputs)
+        val frameUrl = inputs.frameUrl
+        val buttonIndex = inputs.buttonIndex
+        val inputText = inputs.inputText
+        val state = inputs.state
```
nplasterer, line 23:
```suggestion
        val state = inputs.state ?: ""
```

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.xmtp.proto.message.contents.Frames.FrameAction
+import java.util.Date
+
+class FramesClient(private val xmtpClient: Client, var proxy: OpenFramesProxy = OpenFramesProxy()) {
+
+    suspend fun signFrameAction(inputs: FrameActionInputs): FramePostPayload {
+        val opaqueConversationIdentifier = buildOpaqueIdentifier(inputs)
+        val frameUrl = inputs.frameUrl
+        val buttonIndex = inputs.buttonIndex
+        val inputText = inputs.inputText
```
nplasterer, line 22:
```suggestion
        val inputText = inputs.inputText ?: ""
```

Just to match iOS and remove the null check inline

PR #215, opened 2024-04-03
Convo List Performance

nplasterer:
Outputs
```
Loaded 2015 conversations in 17.609s
Second time loaded 2015 conversations in 0.197s
Loaded 2015 conversations in 0.326s
Second time loaded 2015 conversations in 0.192s
```


nplasterer:
Turned out to just be an iOS bug.

PR #217, opened 2024-04-06
Group Message Kind

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/522

Adds the message kind to returned message. Then filters for them in the SDK so that `GroupMembershipChange` Codecs will only get through the messages list if it also has a `Membership_Change` kind.


alexrisch (approved):

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -576,7 +578,10 @@ data class Conversations(
     fun streamAllGroupMessages(): Flow<DecodedMessage> = callbackFlow {
         val messageCallback = object : FfiMessageCallback {
             override fun onMessage(message: FfiMessage) {
-                trySend(MessageV3(client, message).decode())
+                val decodedMessage = MessageV3(client, message).decode()
+                if (!(decodedMessage.encodedContent.type == ContentTypeGroupMembershipChange && decodedMessage.kind != MessageKind.MEMBERSHIP_CHANGE)) {
+                    trySend(decodedMessage)
+                }
```
richardhuaaa, line 584: Would it make sense to move the filtering logic into `decode()`? Then it automatically applies to all call-sites, and can't be screwed up in future call-sites either

cameronvoell (approved):
Updates + tests all look good. One thing I noticed is that for me this PR branch shows `LibXMTP Version` empty in the logs for some reason. I'm not sure if maybe bindings were generated in a different way that could cause that. I double tested the version shows up for fine me on `main`, but not this branch for some reason.

```
LibXMTP Version: 
Branch: 
Date:
```

Separate from the version not showing up in logs for some reason - it still feels to me that the libxmtp version should be stored somewhere in the project so that we don't have to build and check logs to see what commit hash of libxmtp android is using. Maybe until we automate that, just adding the libxmtp commit hash to description of PRs that update the bindings would be a good failsafe. wdyt?

library/src/main/java/org/xmtp/android/library/Conversations.kt
nplasterer: Good call!

richardhuaaa:
It might be nice to ensure that `make libxmtp-version` triggers in `cross_build.sh` instead of `build.rs` and `gen_kotlin.sh`, but the github action would need to be updated too. Hoping I can hand this off to someone?

nplasterer:
I copied over the libxmtp txt file. Next time I'm in libxmtp I can take a look at updating the script. 👀

PR #219, opened 2024-04-09
feat: Rust V2

nplasterer:
Part of: https://github.com/xmtp/xmtp-android/issues/223
We've been seeing some android GRPC issues and also discrepancies between how Android and iOS work. Lets make them more inline with each other by both using the V2 Rust functions.


library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -153,47 +122,125 @@ data class GRPCApiClient(
...
+    override suspend fun publish(envelopes: List<Envelope>) {
+        val ffiEnvelopes = envelopes.map { envelopeToFFi(it) }
+        val request = FfiPublishRequest(envelopes = ffiEnvelopes)
+
+        rustV2Client.publish(request = request, authToken = authToken ?: "")
+    }
+
+    override suspend fun subscribe(topics: List<String>): Flow<Envelope> = flow {
+        try {
+            val subscription = rustV2Client.subscribe(FfiV2SubscribeRequest(topics))
```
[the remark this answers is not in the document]
nplasterer: I toyed with the idea of spinning up a Android GRPC channel just for the subscribe stuff since it feels better than this rust subscribe work. But thought it would be better to just remove the dependency on android grpc entirely. Curious others thoughts.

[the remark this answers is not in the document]
neekolas: That's a little scary. Have you been able to test this with the example app to see if it's the tests or the implementation?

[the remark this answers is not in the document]
nplasterer: Ya I ran into this with Groups as well. Everything seems to work when not using the turbine coroutine test methods. But it does make me desire a beta branch for changes like this.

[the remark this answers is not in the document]
nplasterer: And yes the example app works with streaming.

nplasterer:
I didn't feel comfortable with the major overhaul of streaming here. So decided to bring back the grpc stuff just for streaming. I still think this is an improvement and have confirmed streams still work as expected. I think I will merge this as an incremental step and open another PR for just the subscribe bits so that we can test them in isolation.

example/src/main/java/org/xmtp/android/example/MainActivity.kt
```diff
@@ -182,6 +183,7 @@ class MainActivity : AppCompatActivity(),
 
     private fun showError(message: String) {
         val error = message.ifBlank { resources.getString(R.string.error) }
+        Log.e("MainActivity", message)
```
alexrisch, line 186: Want this log still?

library/src/androidTest/java/org/xmtp/android/library/ContactsTest.kt
```diff
@@ -27,19 +27,6 @@ class ContactsTest {
         assertEquals(contactBundle?.walletAddress, fixtures.bob.walletAddress)
     }
 
-    @Test
-    fun testCachesContacts() {
```
alexrisch, line 31: Any reason this was removed?

alexrisch (approved):

library/src/androidTest/java/org/xmtp/android/library/ContactsTest.kt
nplasterer: Ya I removed all the FakeClient stuff because it's not very useful and is Fake so doesn't really test real world application.  Line 36 and 33 are both things that are from that FakeClient.

example/src/main/java/org/xmtp/android/example/MainActivity.kt
nplasterer: Ya it helps with logging errors in the example. Otherwise it just says "An Error Happened" with no details.

library/src/main/java/README.md
```diff
@@ -2,14 +2,15 @@
...
-1. From repo [libxmtp](https://github.com/xmtp/libxmtp) checkout the branch you would like to make a release from
-2. Navigate to the `bindings_ffi` folder
-3. Follow the instructions for "Rebuilding this crate in the `bindings_ffi` [README](https://github.com/xmtp/libxmtp/tree/main/bindings_ffi#rebuilding-this-crate)"
-4. Copy the contents of `libxmtp/bindings_ffi/src/uniffi/xmtpv3/xmtpv3.kt` to `xmtp-android/library/src/main/java/xmtpv3.kt`
-5. Run format (cmd + opt + l) function to keep the code format consistent and diff small
-6. All instances of `value.forEach` should be changed to `value.iterator().forEach` to be compatible with API 23
-7. Copy the jniLibs from `libxmtp/bindings_ffi/jniLibs` to `xmtp-android/library/src/main/jniLibs`
+## Process for updating from a [libxmtp](https://github.com/xmtp/libxmtp) Kotlin Binding Release
+
+1. From repo [libxmtp](https://github.com/xmtp/libxmtp) run the [kotlin release action](https://github.com/xmtp/libxmtp/actions/workflows/release-kotlin-bindings.yml) for the branch you desire
```
cameronvoell, line 7: nice excited to try this!

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -152,10 +153,14 @@ class Client() {
 
         fun canMessage(peerAddress: String, options: ClientOptions? = null): Boolean {
             val clientOptions = options ?: ClientOptions()
-            val api = GRPCApiClient(
-                environment = clientOptions.api.env,
-                secure = clientOptions.api.isSecure,
-            )
+            val v2Client = runBlocking {
```
cameronvoell, line 156: question - is there a reason we use `runBlocking` instead of making canMessage a `suspend` function? I know we previously had more runBlocking, but we I think we pulled them out because we thought we had better performance with using suspend functions in some cases. same would apply to the other runBlocking in this class.

cameronvoell (approved):
Looks good, just left a comment that I think switching some of the runBlocking for suspend functions could improve performance in some cases.

PR #224, opened 2024-04-12
Bump consent pagination

nplasterer:
This bumps the consent record pagination size to 500 instead of the previous 100. Should help with performance of the subscribe feature.

Previously
```
Loaded 7561 consent envelopes in 8.542s
Loaded 1102 consent entries in 2.429s
Second time loaded 0 consent envelopes in 0.09s
Second time loaded 1102 consent entries in 0s
```

Now
```
Loaded 7561 consent envelopes in 5.011s
Loaded 1102 consent entries in 2.264s
Second time loaded 0 consent envelopes in 0.102s
Second time loaded 1102 consent entries in 0s
```

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/messages/PagingInfo.kt
+++ b/library/src/main/java/org/xmtp/android/library/messages/PagingInfo.kt
@@ -16,35 +16,13 @@ data class Pagination(
 ) {
     val pagingInfo: PagingInfo
         get() {
-            return PagingInfo.newBuilder().also {
-                if (limit != null) {
-                    it.limit = limit
+            return PagingInfo.newBuilder().also { page ->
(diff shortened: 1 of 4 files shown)


library/src/main/java/org/xmtp/android/library/messages/PagingInfo.kt
```diff
@@ -16,35 +16,13 @@ data class Pagination(
...
-                if (cursor != null) {
-                    it.cursor = cursor
-                }
-                if (direction != null) {
-                    it.direction = direction
-                }
-            }.build()
-        }
-    }
-}
```
nplasterer, line 50: This was dead code

alexrisch (approved):

Issue #223, opened 2024-04-12
Bug: GRPC sometimes errors


nplasterer:
### Describe the bug

Some users are seeing this error:
``` 
Failed to load Inbox: Call to function 'XMTP.listConversations' has been rejected.→ Caused by: io.grpc.StatusException: UNAVAILABLE: Unable to resolve host production.xmtp.network 
```

neekolas:
That's starting to look more like an issue on the nodes. Are you able to reproduce the same error when using `grpc.production.xmtp.network`?

neekolas:
I did a little digging in to the metrics from AWS, and there are some spikes of errors coming from the load balancer. These are the `ELB_5XX`s metric over the past week.

![CleanShot 2024-04-12 at 13 31 53@2x](https://github.com/xmtp/xmtp-android/assets/65710/e93a7093-308a-42c3-9a71-f221369fab9c)

Amazon describes that metric like this, which rules out issues on our nodes. It seems like the Amazon Managed LB is failing on its own (maybe through slow autoscaling or an Amazon incident)

> The number of HTTP 5XX server error codes that originate from the load balancer. This count does not include any response codes generated by the targets.

neekolas:
The metrics from the node are healthy, but that would be consistent with a load balancer error since the requests never even would make it to the node.

nplasterer:
The android repo does use `grpc.production.xmtp.network` there is nothing in the android repo that actually chomps the grpc portion of the host.... I looked a little into the grpc android library and I think it's trying to treat grpc as the scheme? When I add https it fails with a no scheme error for https.

Fwiw I was able to reproduce this once. But after restarting my computer and updating android studio I haven't been able to. We do know atleast 3 others are experiencing this on there actual device though.

PR #226, opened 2024-04-13
Increase network timeouts and retries

nplasterer:
Closes https://github.com/xmtp/xmtp-android/issues/223

I'm seeing the tests pass now even on poor connection.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -183,34 +183,7 @@ class Client() {
         options: ClientOptions? = null,
         account: SigningKey? = null,
     ): Client {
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -183,34 +183,7 @@ class Client() {
...
-
-        return Client(
-            address = address,
-            privateKeyBundleV1 = bundle,
-            apiClient = apiClient,
-            libXMTPClient = v3Client,
-            dbPath = dbPath,
-            installationId = v3Client?.installationId()?.toHex() ?: ""
-        )
+        return buildFromV1Bundle(bundle, options, account)
```
nplasterer, line 186: This was duplicate code inside of `buildFromV1Bundle` so instead of trying to keep to copies of the code just call into the other one.

neekolas (approved):

PR #227, opened 2024-04-16
fix: Performance Issues on Buffer allocation

alexrisch:
Fix buffer allocation/deallocation causing slow downs
Added buffer pool and ability to pass buffer

### DIFF
--- a/library/src/main/java/xmtpv3.kt
+++ b/library/src/main/java/xmtpv3.kt
@@ -17,6 +17,7 @@ package uniffi.xmtpv3;
 // compile the Rust component. The easiest way to ensure this is to bundle the Kotlin
 // helpers directly inline like we're doing here.
 
+import RustBufferPool
 import com.sun.jna.Library
 import com.sun.jna.IntegerType
 import com.sun.jna.Native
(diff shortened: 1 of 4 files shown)


library/src/main/java/xmtpv3.kt
```diff
@@ -3935,11 +4160,22 @@ fun `sha256`(`input`: ByteArray): ByteArray {
...
+    val pub = FfiConverterByteArray.lower(`publicKey`, pubBuffer)
+    val priv = FfiConverterByteArray.lower(`privateKey`)
+    val mess = FfiConverterByteArray.lower(`message`)
+    println("here1115 before call")
+    val data =
         rustCallWithError(GenericException) { _status ->
-            _UniFFILib.INSTANCE.uniffi_xmtpv3_fn_func_user_preferences_decrypt(FfiConverterByteArray.lower(`publicKey`),FfiConverterByteArray.lower(`privateKey`),FfiConverterByteArray.lower(`message`),_status)
-        })
+            _UniFFILib.INSTANCE.uniffi_xmtpv3_fn_func_user_preferences_decrypt(pub,priv,mess,_status)
+        }
```
alexrisch, line 4174: When buffer pool is being used it's not getting past here

PR #228, opened 2024-04-16
Fix stream timeouts

nplasterer:
Make sure that streams don't disconnect after 2 minutes, 5 minutes, and 10 minutes. Also confirmed that the poor network tests still work.

Was seeing `io.grpc.okhttp.OkHttpClientTransport$ClientFrameHandler@8913bec: Received GOAWAY with ENHANCE_YOUR_CALM. Debug data: too_many_pings` but no longer with 3 minute for timeout.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/ApiClient.kt
+++ b/library/src/main/java/org/xmtp/android/library/ApiClient.kt
@@ -95,7 +95,6 @@ data class GRPCApiClient(
                     "backoffMultiplier" to 2.0,
                     "retryableStatusCodes" to listOf(
                         "UNAVAILABLE",
-                        "CANCELLED",
                     )
                 )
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/ApiClient.kt
```diff
@@ -95,7 +95,6 @@ data class GRPCApiClient(
                     "backoffMultiplier" to 2.0,
                     "retryableStatusCodes" to listOf(
                         "UNAVAILABLE",
-                        "CANCELLED",
```
nplasterer, line 98: O don't think we should retry if it's canceled. That's likely an action taken on purpose.

alexrisch (approved):

PR #238, opened 2024-04-24
Fix CI Integration Workflow

zombieobject:
This PR is intended to fix the Integration CI workflow. The libxmtp binaries have been updated and failing tests have been marked `@Ignored` for now to focus on the failure point which was related to the presence of dynamically linked libatomic  symbols as noted in the following libxmtp PR: https://github.com/xmtp/libxmtp/pull/709

### DIFF
--- a/library/src/androidTest/java/org/xmtp/android/library/CodecTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/CodecTest.kt
@@ -77,7 +77,7 @@ class CodecTest {
     }
 
     @Test
-    @Ignore("EM: Temporary ignore for failing test while fixing CI")
+    @Ignore("Flaky: CI")
     fun testCanGetPushInfoBeforeDecoded() {
         val codec = NumberCodec()
         Client.register(codec = codec)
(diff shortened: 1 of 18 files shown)


library/src/androidTest/java/org/xmtp/android/library/ReactionTest.kt
```diff
@@ -101,6 +102,7 @@ class ReactionTest {
     }
 
     @Test
+    @Ignore("EM: Temporary ignore for failing test while fixing CI")
     fun testShouldPushMustBeTrue() {
```
nplasterer, line 106: Ya this one is flaky. Might be okay to just ignore because of flaky.

library/src/androidTest/java/org/xmtp/android/library/CodecTest.kt
```diff
@@ -76,6 +77,7 @@ class CodecTest {
     }
 
     @Test
+    @Ignore("EM: Temporary ignore for failing test while fixing CI")
     fun testCanGetPushInfoBeforeDecoded() {
```
nplasterer, line 81: This one is also known to be flaky also probably fine to just ignore

nplasterer:
Do you mind updating this text file so we know which version of the binarys this is https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/libxmtp-version.txt

nplasterer (approved):

PR #240, opened 2024-05-04
fix: Corrected key validation

alexrisch:
Corrected key validation and added additional rules

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/KeyUtil.kt
+++ b/library/src/main/java/org/xmtp/android/library/KeyUtil.kt
@@ -1,14 +1,15 @@
 package org.xmtp.android.library
 
+import android.util.Log
 import com.google.protobuf.kotlin.toByteString
+import org.web3j.utils.Numeric
 import org.web3j.crypto.ECDSASignature
 import org.web3j.crypto.Keys
 import org.web3j.crypto.Sign
 import org.web3j.crypto.Sign.SignatureData
-import org.web3j.utils.Numeric
 import org.xmtp.android.library.messages.Signature
 import org.xmtp.android.library.messages.consentProofText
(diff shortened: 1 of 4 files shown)


library/src/main/java/org/xmtp/android/library/KeyUtil.kt
```diff
@@ -82,11 +81,22 @@ object KeyUtil {
     }
 
     fun validateConsentSignature(signature: String, clientAddress: String, peerAddress: String, timestamp: Long): Boolean {
-        val messageData = Signature.newBuilder().build().consentProofText(peerAddress, timestamp).toByteArray()
-        val signatureData = Numeric.hexStringToByteArray(signature)
-        val sig = Signature.parseFrom(signatureData)
-        val recoveredPublicKey = recoverPublicKeyKeccak256(sig.rawData.toByteString().toByteArray(), Util.keccak256(messageData))
-            ?: return false
-        return clientAddress == publicKeyToAddress(recoveredPublicKey)
+        if (timestamp > System.currentTimeMillis()) {
```
nplasterer, line 84: Everywhere else in the code we use
```suggestion
        if (timestamp > Date().time) {
```

library/src/main/java/org/xmtp/android/library/KeyUtil.kt
```diff
@@ -82,11 +81,22 @@ object KeyUtil {
...
+            return false
+        }
+        val thirtyDaysAgo = System.currentTimeMillis() - 30L * 24 * 60 * 60 * 1000
```
nplasterer, line 87:
```suggestion
        val thirtyDaysAgo = Date().time - 30L * 24 * 60 * 60 * 1000
```

nplasterer (approved):

Issue #233, opened 2024-04-19
Bug: StatusException: UNAVAILABLE: End of stream or IOException - the library function should throw exception so we can handle that in our code and do the appropriate action for that


rsappelit:
### Describe the bug

**Point 1:** We have used this code to get the streamMessages of converstations

````
var client = XMTPManager.fetchClient()
            if (client == null) {
                viewModelScope.launch {
                    XMTPManager.setupXMTP {
                        client = XMTPManager.fetchClient()
                        Unit
                    }
                }
            }
            client
                ?.conversations
                ?.streamAllMessages()
                ?.flowWhileShared(
                    subscriptionCount, SharingStarted.WhileSubscribed(1000L)
                )
                ?.flowOn(Dispatchers.IO)
                ?.distinctUntilChanged()
                ?.mapLatest { conversation ->
                    XMTPManager.parseMessage(conversation)
//                    throw Exception("Test Crash")
                }
                ?.catch {
                    Logger.e(TAG, it.message, it)
                    emptyFlow<MessageInfo>()
                } ?: emptyFlow<MessageInfo>()
````

**Point 2 :** If Conversation screen is Active for 15 to 20 minutes then this crash happens

````
FATAL EXCEPTION: DefaultDispatcher-worker-5
                 Process: com.qriptos.app.dev, PID: 26058
                 io.grpc.StatusException: UNAVAILABLE: End of stream or IOException
                 	at io.grpc.Status.asException(Status.java:548)
                 	at io.grpc.kotlin.ClientCalls$rpcImpl$1$1$1.onClose(ClientCalls.kt:300)
                 	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:574)
                 	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:72)
                 	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:742)
                 	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:723)
                 	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
                 	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
                 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
[...]

PR #243, opened 2024-05-13
Surface streaming errors to be caught

nplasterer:
Closes https://github.com/xmtp/xmtp-android/issues/233

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -673,14 +673,6 @@ data class Conversations(
                 }
             } catch (error: CancellationException) {
                 break
-            } catch (error: StatusException) {
-                if (error.status.code == io.grpc.Status.Code.UNAVAILABLE) {
-                    continue
-                } else {
-                    break
-                }
-            } catch (error: Exception) {
-                continue
             }
         }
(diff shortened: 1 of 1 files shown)


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -745,14 +737,6 @@ data class Conversations(
...
                 break
-            } catch (error: StatusException) {
-                if (error.status.code == io.grpc.Status.Code.UNAVAILABLE) {
-                    continue
-                } else {
-                    break
-                }
-            } catch (error: Exception) {
-                continue
             }
```
nplasterer, line 740: @rsappelit is this kind of what you had in mind? Instead of catching these StatusExceptions just let them be thrown so you can catch and handle them?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -745,14 +737,6 @@ data class Conversations(
                 }
             } catch (error: CancellationException) {
                 break
-            } catch (error: StatusException) {
```
neekolas, line 748: Shouldn't we be trying to reconnect when these errors happen. If we throw we leave it up to the app to handle reconnects

PR #246, opened 2024-05-22
feat: New V3 Identity

nplasterer:
Uses the new V3 Identity for Groups. Gets rid of `accountAddress` in favor of `inboxId`.
Will add support for smart contract wallets in a follow up PR.

cc @neekolas @cameronvoell


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -300,7 +310,16 @@ class Client() {
         var dbPath = ""
         val v3Client: FfiXmtpClient? =
             if (isAlphaMlsEnabled(options)) {
-                val alias = "xmtp-${options!!.api.env}-${accountAddress.lowercase()}"
+                var inboxId = getInboxIdForAddress(
```
nplasterer, line 313: Should `createClient` take this new `inboxId`? It appears the signature for createClient actually didn't change?

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -357,14 +376,15 @@ class Client() {
...
-                    v3Client.registerIdentity(account.sign(it)?.rawData)
+            v3Client.signatureRequest()?.let { signatureRequest ->
+                if (account != null) {
+                    account.sign(signatureRequest.signatureText())?.let {
+                        signatureRequest.addEcdsaSignature(it.rawData)
+                    }
+                    v3Client.registerIdentity(signatureRequest)
+                } else {
+                    throw XMTPException("No signer passed but signer was required.")
                 }
```
nplasterer, line 387: I'm not totally sure on how this is supposed to work now. Would love some guidance. 🙏

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -185,13 +186,13 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         }
     }
 
-    fun memberAddresses(): List<String> {
-        return libXMTPGroup.listMembers().map { it.accountAddress }
+    fun memberInboxIds(): List<String> {
+        return libXMTPGroup.listMembers().map { it.inboxId }
```
nplasterer, line 190: Should we have a convenience method thats like `inboxId` to `accountAddress`?

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: I'm getting this error from libxmtp `org.xmtp.android.library.XMTPException: Error creating client Unknown signer`

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -278,13 +278,13 @@ class GroupTest {
         val group = runBlocking {
             alixClient.conversations.newGroup(
                 listOf(
-                    bo.walletAddress,
+                    boClient.inboxId,
                 )
             )
```
nplasterer, line 283: A new group should take a list of inboxIds now correct? Instead of addresses?

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: `createClient` still takes the address of the wallet. The only big change is the dbPath, which should be something like `$inbox_id-$env`

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -300,7 +310,16 @@ class Client() {
...
             if (isAlphaMlsEnabled(options)) {
-                val alias = "xmtp-${options!!.api.env}-${accountAddress.lowercase()}"
+                var inboxId = getInboxIdForAddress(
+                    logger = logger,
+                    host = options!!.api.env.getUrl(),
+                    isSecure = options.api.isSecure,
+                    accountAddress = accountAddress
+                )
+                if (inboxId.isNullOrBlank()) {
+                    inboxId = generateInboxId(accountAddress, SecureRandom().nextLong().toULong())
```
neekolas, line 320: You can just use `0` for the nonce here

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -105,23 +105,23 @@ data class Conversations(
     }
 
     suspend fun newGroup(
-        accountAddresses: List<String>,
+        inboxIds: List<String>,
```
neekolas, line 108: This actually still takes addresses (and then we resolve them to `inbox_id`s internally). If you think it'd be useful we can also give you a version that takes `inbox_id`s.

library/src/main/java/org/xmtp/android/library/Group.kt
neekolas: Each member in `listMembers()` includes a list of `accountAddress`s for that member. So you can use one or both of those.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -233,19 +233,19 @@ class GroupTest {
...
-                    caro.walletAddress
+                    alixClient.inboxId,
+                    caroClient.inboxId
                 )
             )
         }
         runBlocking { alixClient.conversations.syncGroups() }
         val group = runBlocking { alixClient.conversations.listGroups().first() }
-        runBlocking { group.removeMembers(listOf(caro.walletAddress)) }
+        runBlocking { group.removeMembers(listOf(caroClient.inboxId)) }
```
neekolas, line 243: `removeMembers` still takes addresses. We just resolve them to `inbox_id` for you under the hood.

We do have a `remove_members_by_inbox_id` and `add_members_by_inbox_id` method, although now that I think about it we don't expose those via FFI. That's an easy add.

I'd imagine that'd be useful for tests, and probably some actual user flows too.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
neekolas: Still takes addresses. We can give you a method that takes `inbox_id`s as well if you think it's useful. Something like `create_group_with_inboxes(inbox_ids)`

library/src/main/java/org/xmtp/android/library/Group.kt
nplasterer: Is there such a thing as a primary address?

neekolas: Not really

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Should we still allow the integrator to pass a dbPath?

neekolas: What if we let them pass in a `dbFolder`?

library/src/main/java/org/xmtp/android/library/Client.kt
neekolas: I can take a quick look into that. That error is coming from the original `createClient`? Or it's happening down here?

neekolas: I created a [test case](https://github.com/xmtp/libxmtp/pull/767) that reproduces an error with legacy keys. Not sure if that error is the same one you're facing.

library/src/main/java/org/xmtp/android/library/Group.kt
nplasterer: Okay hmmm wondering if I should return a map here of all addresses for each inbox id. instead of call first on the addresses.

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -153,20 +153,21 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
...
-    fun addedByAddress(): String {
-        return libXMTPGroup.addedByAddress()
+    fun addedByInboxId(): String {
+        return libXMTPGroup.addedByInboxId()
     }
 
     fun permissionLevel(): GroupPermissions {
-        return metadata.policyType()
+        TODO()
+//        return metadata.policyType()
```
nplasterer, line 162: @cameronvoell I assume you'll be updating this with your new permissions work so just commenting out as a TODO for now.

cameronvoell: yep, I can fix that in my permissions PR 👍

library/src/main/java/xmtpv3.kt
```diff
@@ -1644,34 +2196,135 @@ inline fun <T : Disposable?, R> T.use(block: (T) -> R) =
...
+        // For safety's sake: if the library hasn't been run in android_cleaner = true
+        // mode, but is being run on Android, then we still need to think about
+        // Android API versions.
+        // So we check if java.lang.ref.Cleaner is there, and use that…
+        java.lang.Class.forName("java.lang.ref.Cleaner")
+        JavaLangRefCleaner()
+    } catch (e: ClassNotFoundException) {
+        // … otherwise, fallback to the JNA cleaner.
+        UniffiJnaCleaner()
+    }
```
nplasterer, line 2262: @cameronvoell maybe it's just a linter bugger seems to have the fallback here...

nplasterer: Which would catch the error if the create() function is called

cameronvoell: yea I think you're right 👍

cameronvoell (approved):
looking good! 🚀

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -162,19 +161,27 @@ class GroupTest {
             )
         }
         assertEquals(
-            group.memberAddresses().sorted(),
+            group.members().map { it.inboxId }.sorted(),
             listOf(
-                caro.walletAddress.lowercase(),
-                alix.walletAddress.lowercase(),
-                bo.walletAddress.lowercase()
+                caroClient.inboxId.lowercase(),
```
neekolas, line 166: Odd that we have to lowercase `inboxId`s here. Since these always originate from inside libxmtp, can't we just leave the casing alone?

nplasterer: Ahh yes no longer needed now that it's not an address

library/lint-baseline.xml
```diff
@@ -0,0 +1,48 @@
...
+    <issue
+        id="SimpleDateFormat"
+        message="To get local formatting use `getDateInstance()`, `getDateTimeInstance()`, or `getTimeInstance()`, or use `new SimpleDateFormat(String template, Locale locale)` with for example `Locale.US` for ASCII dates."
+        errorLine1="    val formatter = SimpleDateFormat(&quot;EEE, dd MMM yyyy HH:mm:ss &apos;GMT&apos;&quot;)"
+        errorLine2="                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/main/java/org/xmtp/android/library/messages/Signature.kt"
+            line="53"
+            column="21"/>
+    </issue>
```
nplasterer, line 46: @alexrisch is there a reason you need to use the simple date format? I'm just leaving it as a lint issue we can ignore but would be good to understand why we need it.

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -184,14 +185,30 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         }
     }
 
-    fun memberAddresses(): List<String> {
-        return libXMTPGroup.listMembers().map { it.accountAddress }
+    suspend fun addMembersByInboxId(inboxIds: List<String>) {
+        try {
+            libXMTPGroup.addMembersByInboxId(inboxIds)
+        } catch (e: Exception) {
+            throw XMTPException("User does not have permissions", e)
```
neekolas, line 192: There are other things that can go wrong here besides permissions checks. For example, invalid inbox_ids. Or the network being down. We should probably make the error message more generic and rely on the exception from libxmtp

library/lint-baseline.xml
alexrisch: No solid reason vs any other way to get the same date format

library/src/main/java/org/xmtp/android/library/Group.kt
nplasterer: I think theres gonna be a big pass with permissions when @cameronvoell brings his admin permissions work over. But I'll make this more generic in the meantime.

library/src/main/java/org/xmtp/android/library/libxmtp/MessageV3.kt
```diff
@@ -24,7 +24,7 @@ data class MessageV3(val client: Client, private val libXMTPMessage: FfiMessage)
         get() = libXMTPMessage.convoId
 
     val senderAddress: String
```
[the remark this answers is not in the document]
nplasterer: Nice catch updating.

PR #249, opened 2024-05-30
Remove ability to set direction of group messages

nplasterer:
This wasn't actually doing it correctly anyways and this will make it easier to decentralize

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Group.kt
+++ b/library/src/main/java/org/xmtp/android/library/Group.kt
@@ -95,10 +95,9 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         limit: Int? = null,
         before: Date? = null,
         after: Date? = null,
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -142,11 +135,6 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         ).mapNotNull {
```
nplasterer, line 135: Should be an easy rip out looks like we weren't even exposing direction in the bindings anyways. @neekolas

neekolas: Wait. This is just querying from the local DB. You can still do that in any order you want.

I only care about removing the ability to query from the network in descending order

nplasterer: Oooo... 🤦‍♀️ Thats all entirely in libxmtp for V3! So no SDK issue here. I'll close this.

PR #250, opened 2024-05-30
Inconsistency with timestamps

nplasterer:
Fixes the date issue https://github.com/xmtp/xmtp-react-native/issues/394 and the membership list.

Also makes some better naming for consent items

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Contacts.kt
+++ b/library/src/main/java/org/xmtp/android/library/Contacts.kt
@@ -303,14 +303,14 @@ data class Contacts(
         consentList.publish(entries)
     }
 
-    suspend fun allowInbox(inboxIds: List<String>) {
+    suspend fun allowInboxes(inboxIds: List<String>) {
         val entries = inboxIds.map {
             consentList.allowInboxId(it)
         }
         consentList.publish(entries)
     }
 
-    suspend fun denyInbox(inboxIds: List<String>) {
+    suspend fun denyInboxes(inboxIds: List<String>) {
         val entries = inboxIds.map {
             consentList.denyInboxId(it)
         }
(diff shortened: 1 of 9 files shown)


library/src/main/java/org/xmtp/android/library/libxmtp/MessageV3.kt
```diff
@@ -80,7 +80,7 @@ data class MessageV3(val client: Client, private val libXMTPMessage: FfiMessage)
             topic = Topic.groupMessage(convoId.toHex()).description,
             encodedContent = decode().encodedContent,
             senderAddress = senderInboxId,
-            sentAt = Date(),
+            sentAt = sentAt,
```
nplasterer, line 83: This was the fix for the timestamp issues 🤦‍♀️

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -304,14 +303,14 @@ data class Contacts(
         consentList.publish(entries)
     }
 
-    suspend fun allowInboxId(inboxIds: List<String>) {
+    suspend fun allowInbox(inboxIds: List<String>) {
```
alexrisch, line 306:
```suggestion
    suspend fun allowInboxes(inboxIds: List<String>) {
```

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -304,14 +303,14 @@ data class Contacts(
...
         val entries = inboxIds.map {
             consentList.allowInboxId(it)
         }
         consentList.publish(entries)
     }
 
-    suspend fun denyInboxIds(inboxIds: List<String>) {
+    suspend fun denyInbox(inboxIds: List<String>) {
```
alexrisch, line 313:
```suggestion
    suspend fun denyInboxes(inboxIds: List<String>) {
```

alexrisch (approved):

library/src/main/java/org/xmtp/android/library/Contacts.kt
nplasterer: Ya this is what I named it in react native. But for some reason didn't make these plurals for any of the other ones (groups and addresses) 😮‍💨

PR #251, opened 2024-06-03
Fix create from key bundle

nplasterer:
Allow bundles to be created without requiring a signer.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -44,7 +44,6 @@ import org.xmtp.proto.message.api.v1.MessageApiOuterClass
 import org.xmtp.proto.message.api.v1.MessageApiOuterClass.BatchQueryResponse
 import org.xmtp.proto.message.api.v1.MessageApiOuterClass.QueryRequest
 import uniffi.xmtpv3.FfiXmtpClient
-import uniffi.xmtpv3.LegacyIdentitySource
 import uniffi.xmtpv3.createClient
 import uniffi.xmtpv3.generateInboxId
 import uniffi.xmtpv3.getInboxIdForAddress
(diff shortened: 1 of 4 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -376,13 +376,15 @@ class Client() {
 
         if (v3Client != null) {
             v3Client.signatureRequest()?.let { signatureRequest ->
-                if (account != null) {
-                    account.sign(signatureRequest.signatureText())?.let {
-                        signatureRequest.addEcdsaSignature(it.rawData)
+                if (!signatureRequest.isReady()) {
```
neekolas, line 379: Have you tried removing this `isReady` check. I'm still unconvinced it actually matters with the way you're using `libxmtp`

neekolas (approved):

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Yep doesn't seem to be needed removed it.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -302,10 +299,10 @@ class Client() {
         account: SigningKey?,
         appContext: Context?,
         privateKeyBundleV1: PrivateKeyBundleV1,
-        legacyIdentitySource: LegacyIdentitySource,
-        accountAddress: String,
+        address: String,
     ): Pair<FfiXmtpClient?, String> {
         var dbPath = ""
+        val accountAddress = address.lowercase()
```
nplasterer, line 305: @neekolas I still think there is probably an improvement on the libxmtp side for normalizing addresses maybe in the `getInboxIdForAddress`

neekolas: For sure

PR #262, opened 2024-06-24
Add new fetch methods for V3 local DB

nplasterer:
- adds ability to fetch message by id
- adds ability to fetch group by id
- adds ability to fetch inboxId by address
- adds ability to pass historySyncURL in on create
- adds ability to requestMessageHistorySync

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -17,6 +17,7 @@ import org.web3j.crypto.Keys.toChecksumAddress
 import org.xmtp.android.library.GRPCApiClient.Companion.makeSubscribeRequest
 import org.xmtp.android.library.codecs.ContentCodec
 import org.xmtp.android.library.codecs.TextCodec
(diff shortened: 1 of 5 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -368,7 +370,8 @@ class Client() {
                     accountAddress = accountAddress,
                     inboxId = inboxId,
                     nonce = 0.toULong(),
-                    legacySignedPrivateKeyProto = privateKeyBundleV1.toV2().identityKey.toByteArray()
+                    legacySignedPrivateKeyProto = privateKeyBundleV1.toV2().identityKey.toByteArray(),
+                    historySyncUrl = options.historySyncUrl
```
nplasterer, line 374: @jhaaaa just putting the message history stuff on your radar for documentation. Probably should wait till @tuddman gets back from vacation but we could start spiking it out in the mean time.

cameronvoell (approved):

PR #263, opened 2024-06-25
Adds custom policy type, fixes invalid policy error

cameronvoell:
Related to https://github.com/xmtp/libxmtp/pull/867

### DIFF
--- a/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
@@ -88,18 +88,14 @@ class GroupTest {
         assertEquals(alixGroup.members().size, 3)
         assertEquals(boGroup.members().size, 3)
 
-        runBlocking {
-            alixGroup.removeMembers(listOf(caro.walletAddress))
-            boGroup.sync()
+        // All members also defaults remove to admin only now.
+        assertThrows(XMTPException::class.java) {
+            runBlocking {
+                alixGroup.removeMembers(listOf(caro.walletAddress))
+                boGroup.sync()
+            }
         }
 
-        assertEquals(alixGroup.members().size, 2)
-        assertEquals(boGroup.members().size, 2)
-
-        runBlocking {
-            boGroup.addMembers(listOf(caro.walletAddress))
-            alixGroup.sync()
-        }
         assertEquals(alixGroup.members().size, 3)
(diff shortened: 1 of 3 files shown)


library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -485,7 +484,7 @@ class GroupTest {
 
         Thread.sleep(1000)
 
-        assertEquals(4, messageCallbacks)
+        assertEquals(secondMsgCheck, messageCallbacks)
```
cameronvoell, line 487: this test was failing before the change, think this last value is supposed to also be secondMsgCheck

library/src/main/java/xmtpv3.kt
```diff
@@ -47,17 +47,14 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 open class RustBuffer : Structure() {
     // Note: `capacity` and `len` are actually `ULong` values, but JVM only supports signed values.
     // When dealing with these fields, make sure to call `toULong()`.
-    @JvmField
-    var capacity: Long = 0
-    @JvmField
-    var len: Long = 0
-    @JvmField
-    var data: Pointer? = null
+    @JvmField var capacity: Long = 0
```
nplasterer, line 50: Can you run `command + option + L` on this file to format it. That should bring down the changes.

nplasterer (approved):
Don't forget to copy over this file too https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/libxmtp-version.txt

library/src/main/jniLibs/armeabi-v7a/libuniffi_xmtpv3.so
nplasterer, line 1: I feel mildly concerned that there is only removals on most of these files.

cameronvoell: re-ran crossbuild locally, and had no errors, and files were the same. also tested locally running tests and against RN. Agree getting these built on CI would be a big improvement though 👍

library/src/main/java/xmtpv3.kt
cameronvoell: version file added here 247d07cf8483792097f2017ea9e7f7431ba1e706

PR #275, opened 2024-07-22
Always set an inbox id for V2

nplasterer:
To get React Native working https://github.com/xmtp/xmtp-react-native/pull/440

We need to have the unique identifier of an inboxID set even for V2 clients. This can be done since a v3 client does not need to be created to create an inboxId based on address.


cameronvoell (approved):
looks good!

PR #277, opened 2024-07-23
Allow custom permissions on group creation

cameronvoell:
Related to https://github.com/xmtp/libxmtp/pull/918

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -50,7 +50,6 @@ import uniffi.xmtpv3.FfiV2SubscriptionCallback
 import uniffi.xmtpv3.NoPointer
 import uniffi.xmtpv3.org.xmtp.android.library.libxmtp.GroupPermissionPreconfiguration
 import uniffi.xmtpv3.org.xmtp.android.library.libxmtp.PermissionPolicySet
-import uniffi.xmtpv3.FfiPermissionPolicy
 import uniffi.xmtpv3.FfiGroupPermissionsOptions
 import uniffi.xmtpv3.FfiPermissionPolicySet
 import java.util.Date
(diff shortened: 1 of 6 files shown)


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -135,7 +137,8 @@ data class Conversations(
                     groupName = groupName,
                     groupImageUrlSquare = groupImageUrlSquare,
                     groupDescription = groupDescription,
-                    groupPinnedFrameUrl = groupPinnedFrameUrl
+                    groupPinnedFrameUrl = groupPinnedFrameUrl,
+                    customPermissionPolicySet = if (permissionPolicySet == null) null else PermissionPolicySet.toFfiPermissionPolicySet(permissionPolicySet)
```
nplasterer, line 141: Can you set a permission policy set and a permission like `AdminOnly`? Does the permission policy set just overwrite whatever is passed in for permissions?

nplasterer (approved):

library/src/main/java/org/xmtp/android/library/Conversations.kt
cameronvoell: @nplasterer Have descriptive error messages for if you try and pass in a `permissionPolicySet` with `AdminOnly` or `AllMembers`, see libxmtp code: https://github.com/xmtp/libxmtp/pull/918/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8fR619-R629

Also fails if you try and set it to `CustomPolicy` without passing in a `policySet`. Last test case should show those: https://github.com/xmtp/xmtp-android/pull/277/files#diff-05026c44bd6ddae4119151ab186c979580cae82024649c9e594c78ebab39c826R409-R477 

Figured I'd coordinate with @jhaaaa to get that documented once the changes are merged in all the SDKs.

cameronvoell: We could also just make a separate `newCustomPermissionGroup` function in the SDKs so that it's a little more straightforward. Wasn't sure if that was worthwhile, but happy to update it. wdyt @nplasterer

nplasterer: No I think this makes sense! No need to change.

cameronvoell: ended up adding `newGroupCustomPermissions` here https://github.com/xmtp/xmtp-android/pull/277/commits/7191b552ead625a7cc7094d1ccf2fbb0f0f78c56. think it's worth it just to make the docs etc more straightforward.

PR #283, opened 2024-08-07
update libxmtp for epoch and members fixes

cameronvoell:


nplasterer (approved):

alexrisch (approved):

library/src/main/java/xmtpv3.kt
```diff
@@ -376,11 +380,11 @@ private fun findLibraryName(componentName: String): String {
     if (libOverride != null) {
         return libOverride
     }
-    return "uniffi_xmtpv3"
+    return "xmtpv3"
```
nplasterer, line 383: Seems like with the new uniffi version we use in rust this keeps changing this to remove the uniffi and that breaks everything inside of android. I think I'd like to look more into why this breaks android. But Just going to undo this change for now to get a release out.

PR #284, opened 2024-08-07
feat: Update the bindings and add the default message sync url

nplasterer:
This adds a default message sync url and bumps to the latest rust.

### DIFF
--- a/library/src/main/java/README.md
+++ b/library/src/main/java/README.md
@@ -12,6 +12,7 @@ Kotlin code emitted by the `bindings_ffi` crate in [libxmtp](https://github.com/
 4. Navigate to the [latest release](https://github.com/xmtp/libxmtp/releases) once the action completes
 5. Download the `LibXMTPKotlinFFI.zip` assets
 6. Unzip and then copy the jniLibs to `xmtp-android/library/src/main/jniLibs`
-7. NOTE: sometimes the library name gets updated and will cause failures the library name must be `uniffi_xmtpv3` not `xmtpv3` -- search for `findLibraryName` and make sure the correct name is returned.
+7. All instances of `value.forEach` should be changed to `value.iterator().forEach` to be compatible with API 23
+8. NOTE: sometimes the library name gets updated and will cause failures the library name must be `uniffi_xmtpv3` not `xmtpv3` -- search for `findLibraryName` and make sure the correct name is returned.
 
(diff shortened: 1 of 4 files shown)


library/src/main/java/README.md
```diff
@@ -12,5 +12,6 @@ Kotlin code emitted by the `bindings_ffi` crate in [libxmtp](https://github.com/
 4. Navigate to the [latest release](https://github.com/xmtp/libxmtp/releases) once the action completes
 5. Download the `LibXMTPKotlinFFI.zip` assets
 6. Unzip and then copy the jniLibs to `xmtp-android/library/src/main/jniLibs`
+7. NOTE: sometimes the library name gets updated and will cause failures the library name must be `uniffi_xmtpv3` not `xmtpv3` -- search for `findLibraryName` and make sure the correct name is returned.
```
nplasterer, line 15: I don't know why when we bumped to uniffi version 28 that this started renaming the library. I did a deep dive today through the docs and don't see any reason why this would be happening https://mozilla.github.io/uniffi-rs/latest/kotlin/configuration.html. This is a work around until we can invest more time in it to figure it out.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -67,7 +67,7 @@ data class ClientOptions(
     val enableV3: Boolean = false,
     val dbDirectory: String? = null,
     val dbEncryptionKey: ByteArray? = null,
-    val historySyncUrl: String? = null,
+    val historySyncUrl: String = "https://message-history.dev.ephemera.network/",
```
alexrisch, line 70: Should this point to prod when set to prod?
Do we need additional checks like local env?

nplasterer: Yep already fixed. 👍

library/src/main/java/README.md
cameronvoell: Re- lint failure, not sure if we should maybe re-add this line as well:

https://github.com/xmtp/xmtp-android/blame/9e47f24fd3b714be09e6fedfc7364c97a68d0c8a/library/src/main/java/README.md#L12

nplasterer: Good call that one came back too. 😭

neekolas (approved):

PR #297, opened 2024-09-13
Consent V3

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/801

This adds the ability to get and set consent records in V3

Dual sends so we are always setting both V2 and V3 consent. And adds a check to see if a V2Client is present to support pure V3 clients.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -90,10 +90,9 @@ class Client() {
     var logger: XMTPLogger = XMTPLogger()
     val libXMTPVersion: String = getVersionInfo()
     var installationId: String = ""
-    private var v3Client: FfiXmtpClient? = null
+    var v3Client: FfiXmtpClient? = null
     var dbPath: String = ""
     lateinit var inboxId: String
-    var hasV3Client: Boolean = false
     var hasV2Client: Boolean = false
 
     companion object {
@@ -359,7 +358,6 @@ class Client() {
             }
 
         if (v3Client != null) {
-            hasV3Client = true
             options.preAuthenticateToInboxCallback?.let {
                 runBlocking {
                     it.invoke()
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -93,7 +93,7 @@ class Client() {
     var v3Client: FfiXmtpClient? = null
     var dbPath: String = ""
     lateinit var inboxId: String
-    var hasV2Client: Boolean = false
+    var hasV2Client: Boolean = true
 
     companion object {
         private const val TAG = "Client"
@@ -278,7 +278,6 @@ class Client() {
         options: ClientOptions? = null,
         account: SigningKey? = null,
     ): Client {
-        hasV2Client = true
         val address = v1Bundle.identityKey.publicKey.recoverWalletSignerPublicKey().walletAddress
         val newOptions = options ?: ClientOptions()
         val v2Client =
@@ -397,7 +396,6 @@ class Client() {
         apiClient: ApiClient,
         options: ClientOptions? = null,
     ): PrivateKeyBundleV1 {
-        hasV2Client = true
         val keys = loadPrivateKeys(account, apiClient, options)
         return if (keys != null) {
             keys
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -93,6 +93,8 @@ class Client() {
     private var v3Client: FfiXmtpClient? = null
     var dbPath: String = ""
     lateinit var inboxId: String
+    var hasV3Client: Boolean = false
+    var hasV2Client: Boolean = false
 
     companion object {
         private const val TAG = "Client"
@@ -277,6 +279,7 @@ class Client() {
         options: ClientOptions? = null,
         account: SigningKey? = null,
     ): Client {
+        hasV2Client = true
         val address = v1Bundle.identityKey.publicKey.recoverWalletSignerPublicKey().walletAddress
         val newOptions = options ?: ClientOptions()
         val v2Client =
@@ -356,6 +359,7 @@ class Client() {
             }
 
         if (v3Client != null) {
+            hasV3Client = true
             options.preAuthenticateToInboxCallback?.let {
                 runBlocking {
                     it.invoke()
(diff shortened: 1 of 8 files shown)


library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -125,80 +165,95 @@ class ConsentList(
...
-                                it.setDenyAddress(
-                                    PrivatePreferencesAction.DenyAddress.newBuilder()
-                                        .addWalletAddresses(entry.value),
-                                )
-
-                            ConsentState.UNKNOWN -> it.clearMessageType()
+        if (client.v3Client != null) {
+            setV3ConsentState(entries)
+        }
+        if (client.hasV2Client) {
```
nplasterer, line 171: This is in prep for when we have SCW. Only publish if we have a V2 client.

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -125,80 +165,95 @@ class ConsentList(
...
 
-        client.publish(listOf(envelope))
+    suspend fun setV3ConsentState(entries: List<ConsentListEntry>) {
+        entries.forEach {
+            client.v3Client?.setConsentState(
+                ConsentState.toFfiConsentState(it.consentType),
+                EntryType.toFfiConsentEntityType(it.entryType),
+                it.value
+            )
+        }
```
nplasterer, line 256: For performance reasons in this interim period of V2 and V3 we might want to add a FFibinding for setting a list so we dont have to iterate.

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -270,7 +346,8 @@ data class Contacts(
 ) {
 
     suspend fun refreshConsentList(): ConsentList {
-        consentList.load()
+        val entries = consentList.load()
+        consentList.setV3ConsentState(entries)
```
nplasterer, line 350: After loading the V2 consent add them all to the V3 database.

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -243,21 +298,42 @@ class ConsentList(
...
+            return ConsentState.fromFfiConsentState(
+                it.getConsentState(
+                    FfiConsentEntityType.ADDRESS,
+                    address
+                )
+            )
+        }
         val entry = entries[ConsentListEntry.address(address).key]
-
         return entry?.consentType ?: ConsentState.UNKNOWN
```
nplasterer, line 311: Previously it was checking this local entries list which we could still just do since we write to both. But I like move towards checking the V3 state only since that should be the ultimate source of truth if you have V3.

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -178,6 +178,15 @@ class Group(val client: Client, private val libXMTPGroup: FfiGroup) {
         return MessageV3(client, message)
     }
 
+    fun updateConsentState(state: ConsentState) {
+        val consentState = ConsentState.toFfiConsentState(state)
+        libXMTPGroup.updateConsentState(consentState)
```
nplasterer, line 183: While we still dont have history we should probably update the V2 one here as well 🤔

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -90,9 +90,10 @@ class Client() {
     var logger: XMTPLogger = XMTPLogger()
     val libXMTPVersion: String = getVersionInfo()
     var installationId: String = ""
-    private var v3Client: FfiXmtpClient? = null
+    var v3Client: FfiXmtpClient? = null
     var dbPath: String = ""
     lateinit var inboxId: String
+    var hasV2Client: Boolean = false
```
cameronvoell, line 96: Just a nit - looks like it is currently impossible to create a `Client` without `hasV2Client` being `true`, but it is pretty hard to tell that from this code. (you need to visually verify that all functions that return Client either call `buildFromV1Bundle` or `loadOrCreateKeys`)

Since currently all code paths set this to `true`, I wonder if we should just set it to true by default, instead of false, and add the hasV2Client = `false` later on when there is a path that actually makes this possible.

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -125,80 +163,95 @@ class ConsentList(
...
+                                    )
+
+                                ConsentState.UNKNOWN -> it.clearMessageType()
+                            }
                         }
                     }
                 }
-            }
-        }.build()
+            }.build()
```
cameronvoell, line 228: cool trick, in kotlin you can use `to` infix notation to create `Pair`s to make this more concise:

```kotlin
when (entry.entryType to entry.consentType) {
                        EntryType.ADDRESS to ConsentState.ALLOWED -> setAllowAddress(
                            PrivatePreferencesAction.AllowAddress.newBuilder()
                                .addWalletAddresses(entry.value)
                        )

                        EntryType.ADDRESS to ConsentState.DENIED -> setDenyAddress(
                            PrivatePreferencesAction.DenyAddress.newBuilder()
                                .addWalletAddresses(entry.value)
                        )

                        EntryType.GROUP_ID to ConsentState.ALLOWED -> setAllowGroup(
                            PrivatePreferencesAction.AllowGroup.newBuilder()
                                .addGroupIds(entry.value)
                        )

                        EntryType.GROUP_ID to ConsentState.DENIED -> setDenyGroup(
                            PrivatePreferencesAction.DenyGroup.newBuilder().addGroupIds(entry.value)
                        )

                        EntryType.INBOX_ID to ConsentState.ALLOWED -> setAllowInboxId(
                            PrivatePreferencesAction.AllowInboxId.newBuilder()
                                .addInboxIds(entry.value)
                        )

                        EntryType.INBOX_ID to ConsentState.DENIED -> setDenyInboxId(
                            PrivatePreferencesAction.DenyInboxId.newBuilder()
                                .addInboxIds(entry.value)
                        )

                        else -> clearMessageType()
                    }
                }
            }.build()
```

not that common though so up to you

cameronvoell (approved):
🚀

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Hmmm you know this is a good point. I actually think it's going to be really hard to validate the correctness of all our work without having the ability to make a pure v3 client. I'm going to open a separate PR today that so that we can make certain this stuff is working like expected in V3 and not just a side effect of the dual sending.

nplasterer: From creating this PR I align with you it should be defaulted to true https://github.com/xmtp/xmtp-android/pull/298/files

PR #298, opened 2024-09-16
Create a pure v3 client

nplasterer:
In order to properly test 1to1 it would be nice to be able to create a pure v3 client. This will be necessary once we support SCW as well.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -54,6 +54,8 @@ import java.time.Instant
 import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
+import kotlin.random.Random
+import kotlin.random.nextULong
 
 typealias PublishResponse = org.xmtp.proto.message.api.v1.MessageApiOuterClass.PublishResponse
 typealias QueryResponse = org.xmtp.proto.message.api.v1.MessageApiOuterClass.QueryResponse
@@ -90,9 +92,10 @@ class Client() {
     var logger: XMTPLogger = XMTPLogger()
     val libXMTPVersion: String = getVersionInfo()
     var installationId: String = ""
-    private var v3Client: FfiXmtpClient? = null
+    var v3Client: FfiXmtpClient? = null
     var dbPath: String = ""
     lateinit var inboxId: String
+    var hasV2Client: Boolean = true
 
     companion object {
         private const val TAG = "Client"
@@ -198,6 +201,23 @@ class Client() {
         this.inboxId = inboxId
     }
 
+    constructor(
+        address: String,
+        libXMTPClient: FfiXmtpClient,
+        dbPath: String,
+        installationId: String,
+        inboxId: String,
+    ) : this() {
+        this.address = address
+        this.contacts = Contacts(client = this)
+        this.v3Client = libXMTPClient
+        this.conversations =
+            Conversations(client = this, libXMTPConversations = libXMTPClient.conversations())
+        this.dbPath = dbPath
+        this.installationId = installationId
+        this.inboxId = inboxId
+    }
+
     suspend fun buildFrom(
         bundle: PrivateKeyBundleV1,
         options: ClientOptions? = null,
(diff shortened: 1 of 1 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -338,6 +391,7 @@ class Client() {
 
                 val encryptionKey = options.dbEncryptionKey
                     ?: throw XMTPException("No encryption key passed for the database. Please store and provide a secure encryption key.")
+                val nonce = if (privateKeyBundleV1 != null) 0UL else Random.nextULong()
```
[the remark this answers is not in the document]
neekolas: You can still use nonce 0 for clients without a privateKeyBundle. You just cannot use nonce > 0 for clients created with one.

In either case, I'd be very careful with generating a random nonce as part of building the client. This should only happen if there is no inbox_id associated with the wallet. Otherwise you could create multiple inboxes for the same wallet address.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -533,13 +588,14 @@ class Client() {
     suspend fun publish(envelopes: List<Envelope>) {
         val authorized = AuthorizedIdentity(
             address = address,
-            authorized = privateKeyBundleV1.identityKey.publicKey,
-            identity = privateKeyBundleV1.identityKey,
+            authorized = v1keys.identityKey.publicKey,
+            identity = v1keys.identityKey,
         )
+        val client = apiClient ?: throw XMTPException("V2 only function")
```
cameronvoell, line 594: just a nit, we could check for null `apiClient` first in the function, since there's no need creating AuthorizedIdentity if apiClient is null.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -495,7 +549,8 @@ class Client() {
         request: FfiV2SubscribeRequest,
         callback: FfiV2SubscriptionCallback,
     ): FfiV2Subscription {
-        return apiClient.subscribe(request, callback)
+        val client = apiClient ?: throw XMTPException("V2 only function")
```
cameronvoell, line 552: one style thing in this file, in many existing places in this file, we use the "safe call with let" syntax to check if the v3Client is null. I don't have too strong a preference that syntax vs using the elvis operator like here, but feel a little strange for it to be half one way, half the other. see inboxIdFromAddress below as an example:

```kotlin
suspend fun inboxIdFromAddress(address: String): String? {
        v3Client?.let {
            return it.findInboxId(address.lowercase())
        }
        throw XMTPException("Error no V3 client initialized")
    }
```

nplasterer: I'll try to unify it

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -670,11 +726,11 @@ class Client() {
...
+        get() = PrivateKeyBundleBuilder.buildFromV1Key(v1keys)
 
     val v1keys: PrivateKeyBundleV1
-        get() = privateKeyBundleV1
+        get() = privateKeyBundleV1 ?: throw XMTPException("V2 only function")
 
     val keys: PrivateKeyBundleV2
-        get() = privateKeyBundleV1.toV2()
+        get() = v1keys.toV2()
 }
```
cameronvoell, line 736: logic all seems fine here, but these names seem very confusing for me personally. The names aren't new in this PR though, so no need to block the change. 

an example improvement might be:
( we already have the nullable class field `privateKeyBundleV1`)
- `v1keys` could instead be called something like `nonNullPrivateKeyBundleV1`
- `keys` could instead be called `privateKeyBundleV2`

cameronvoell (approved):
Changes all look good, just adding some ideas/nits related to code organization. I think the new client v3 constructor, and making `privateKeyBundleV1` and `apiClient` both [nullable](https://github.com/xmtp/xmtp-android/pull/298/files#diff-7f5f11ae5a1cc09075430850830b536bb9e472ea3e97397c396cfefea4e1e0daR90-R91) is the right idea 👍

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -54,6 +54,8 @@ import java.time.Instant
 import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
+import kotlin.random.Random
+import kotlin.random.nextULong
```
cameronvoell, line 58: Unused
```suggestion
```

library/src/main/java/org/xmtp/android/library/Client.kt
nplasterer: Ya this one really thew me off as well something I will be happy to kill with V2.

nplasterer:
All tests pass 🎉 

Screenshot 2024-09-18 at 11 48 31 AM

cameronvoell (approved):
Nice updates! :shipit:

PR #304, opened 2024-10-03
Smart contract wallet support

nplasterer:
This adds support for smart contract wallet accounts.


library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -18,13 +18,16 @@ import java.util.Date
 
 interface SigningKey {
     val address: String
-    /// If this signing key is a smart contract wallet
+
+    // If this signing key is a smart contract wallet
```
Koleok, line 22: This was actually stopping the build for me due to a ktLint error, the rest is just auto-formatting

nplasterer: Sorry about that!

library/src/androidTest/java/org/xmtp/android/library/TestHelpers.kt
```diff
@@ -41,6 +52,94 @@ class FakeWallet : SigningKey {
...
+
+    override suspend fun sign(data: ByteArray): Signature {
+        val smartWallet = CoinbaseSmartWallet.deploy(
+            web3j,
+            credentials,
+            DefaultGasProvider()
+        ).send()
+
+        val randomHash = ByteArray(32)
+        SecureRandom().nextBytes(randomHash)
```
neekolas, line 95: Why is the hash random?

This is going through most of the steps to get a real, verifiable, signature...but if it's signing randomness it's not going to verify with a genuine SCW.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -289,34 +289,62 @@ class Client() {
...
         account: SigningKey,
-        options: ClientOptions,
+        options: ClientOptions? = null,
     ): Client {
         this.hasV2Client = false
-        val inboxId = getOrCreateInboxId(options, account.address)
-
+        val clientOptions = options ?: ClientOptions(enableV3 = true)
+        val accountAddress =
+            if (account.isSmartContractWallet) "eip155:${account.chainId}:${account.address.lowercase()}" else account.address.lowercase()
```
nplasterer, line 328: This was the fix for the `Unknown Signer` issue. We need to create the client with the CIAP address.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -289,34 +289,62 @@ class Client() {
...
-                    dbPath,
-                    client.installationId().toHex(),
-                    client.inboxId(),
-                    options.api.env
-                )
-            } ?: throw XMTPException("Error creating V3 client: libXMTPClient is null")
+    // Function to build a V3 client without a signing key (using only address (& chainId for SCW))
+    suspend fun buildV3(
+        address: String,
+        contractChainId: Long? = null,
```
nplasterer, line 339: Do people know what scw stand for? I would be okay with calling this `scwChainId` if we don't think it's confusing.

library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -19,9 +19,31 @@ import java.util.Date
...
+
+    suspend fun sign(data: ByteArray): SignatureOuterClass.Signature? {
+        throw NotImplementedError("sign(ByteArray) is not implemented.")
+    }
+
+    suspend fun sign(message: String): SignatureOuterClass.Signature? {
+        throw NotImplementedError("sign(String) is not implemented.")
+    }
+
+    suspend fun signSmartContract(message: String): ByteArray {
```
nplasterer, line 44: Same statement would be okay with calling this `signSCW` if others think that makes it more clear?

library/src/main/java/org/xmtp/android/library/SigningKey.kt
```diff
@@ -19,9 +19,31 @@ import java.util.Date
 interface SigningKey {
     val address: String
 
-    suspend fun sign(data: ByteArray): SignatureOuterClass.Signature?
+    // If this signing key is a smart contract wallet
+    val isSmartContractWallet: Boolean
+        get() = false
 
-    suspend fun sign(message: String): SignatureOuterClass.Signature?
+    // Default chainId value set to 1
```
neekolas, line 26: I'm not sure about this, but maybe we don't want a default chain ID.

If I'm signing in with an EOA, we have no idea what chain I'm really using. If I'm on Base, it's not really fair to assume that I'm using Ethereum.

It could be safer to have the field default to null, and if you have a null chain ID on a SCW to error out at runtime.

library/src/androidTest/java/org/xmtp/android/library/TestHelpers.kt
```diff
@@ -41,6 +54,82 @@ class FakeWallet : SigningKey {
         get() = privateKey.walletAddress
 }
 
+class FakeSCWWallet : SigningKey {
+    private var web3j: Web3j = Web3j.build(HttpService("http://10.0.2.2:8545"))
+    private val credentials: Credentials =
```
neekolas, line 59: The credentials are a little confusing here. They get used to deploy the SCW factory, deploy the smart contract wallet, and they are the wallet that signs messages.

The first two need to be funded, so they kinda have to be the default anvil keys. The last one can be anything so you can create multiple SCWs in tests and have them talk to each other.

Maybe we can break this up to `contractDeployerCredentials` and `ownerCredentials` or something along those lines.

library/src/androidTest/java/org/xmtp/android/library/TestHelpers.kt
```diff
@@ -41,6 +54,82 @@ class FakeWallet : SigningKey {
...
+        Credentials.create("ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80")
```
neekolas, line 60: Would be nice to make this a named constant like `DEFAULT_ANVIL_CREDENTIALS` so others know where these random numbers came from.

nplasterer:
Feel really good about this. I think it's ready for a final review.

Screenshot 2024-10-19 at 6 46 00 AM

neekolas (approved):

neekolas (approved):

PR #306, opened 2024-10-09
Dual Send 1to1 Conversations in V3

nplasterer:
This PR has evolved a lot and I've ended on the simplest function to get a V2/V3 client pre creating conversations and dual sending messages.

This does not do any listing of V2 and V3 conversations in the same list which is where a lot of the complication arises with handling dual sent messages and requires a V2 database in libxmtp.


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -291,30 +291,32 @@ class Client() {
 
     // This is a V3 only feature
     suspend fun createOrBuild(
-        account: SigningKey,
-        options: ClientOptions,
+        address: String,
+        account: SigningKey? = null,
+        options: ClientOptions? = null,
```
nplasterer, line 296: Since this is createOrBuild we only need the signingKey for the create action and the address for the build action. So lets make this nullable.

library/src/main/java/org/xmtp/android/library/Contacts.kt
```diff
@@ -47,15 +47,15 @@ enum class EntryType {
         fun toFfiConsentEntityType(option: EntryType): FfiConsentEntityType {
             return when (option) {
                 EntryType.ADDRESS -> FfiConsentEntityType.ADDRESS
-                EntryType.GROUP_ID -> FfiConsentEntityType.GROUP_ID
+                EntryType.GROUP_ID -> FfiConsentEntityType.CONVERSATION_ID
```
nplasterer, line 50: I'm going to do the breaking change to rename these to `conversation_id` in a follow up.

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -326,14 +304,159 @@ sealed class Conversation {
             is V1 -> conversationV1.streamDecryptedMessages()
             is V2 -> conversationV2.streamDecryptedMessages()
             is Group -> group.streamDecryptedMessages()
+            is Dm -> dm.streamDecryptedMessages()
+        }
+    }
+
+    // ------- V1 V2 to be deprecated ------
```
nplasterer, line 311: Organized this so all functionality that will no longer be supported is below this line.

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -28,154 +29,130 @@ sealed class Conversation {
...
                 is V1 -> conversationV1.sentAt
                 is V2 -> conversationV2.createdAt
                 is Group -> group.createdAt
+                is Dm -> dm.createdAt
             }
         }
 
-    // This is the address of the peer that I am talking to.
-    val peerAddress: String
+    val name: String
```
nplasterer, line 76: Add all the common interface functionality between groups and dms for ease of access.

library/src/main/java/org/xmtp/android/library/ConversationV1.kt
```diff
@@ -195,6 +195,16 @@ data class ConversationV1(
...
+        if (client.v3Client != null) {
+            try {
+                val dm = client.conversations.findOrCreateDm(peerAddress)
+                prepared.encodedContent?.let {
+                    dm.send(it)
+                }
+            } catch (e: Exception) {
+                // Do nothing if this errors
+            }
+        }
```
nplasterer, line 207: Catch all errors and do nothing if the message can't be sent or dm created. Do our best to create V3 equivalent dms but don't error if it doesn't happen.

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -433,73 +482,236 @@ data class Conversations(
...
+            override fun onConversation(conversation: FfiConversation) {
+                trySend(Conversation.Group(Group(client, conversation)))
+            }
+        }
+        val stream = libXMTPConversations?.streamGroups(groupCallback)
+            ?: throw XMTPException("Client does not support Groups")
+        awaitClose { stream.end() }
+    }
+
+    // ------- V1 V2 to be deprecated ------
```
nplasterer, line 699: Try to make the error messages good and put all soon to be deprecated V2 code below this line.

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -320,84 +296,156 @@ data class Conversations(
...
+        }
+    }
+
+    suspend fun listDms(
+        after: Date? = null,
+        before: Date? = null,
+        limit: Int? = null,
+        order: ConversationOrder = ConversationOrder.CREATED_AT,
+    ): List<Dm> {
+        if (client.hasV2Client) throw XMTPException("Only supported for V3 only clients.")
```
nplasterer, line 333: Don't allow clients that have V2 identities to call this function.

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -320,84 +296,156 @@ data class Conversations(
...
+
+        val groups = ffiGroups.map {
+            Group(client, it)
+        }
+
+        return when (order) {
+            ConversationOrder.LAST_MESSAGE ->
+                groups.sortedByDescending { group ->
+                    group.messages(limit = 1).firstOrNull()?.sent
+                }
```
nplasterer, line 321: Allow ordering by last message to get rid of the need for batch messages? Also open to removing this from here and just letting integrators handle the ordering now that its much more performant?

library/src/main/java/org/xmtp/android/library/Dm.kt
```diff
@@ -0,0 +1,267 @@
...
+        get() = libXMTPGroup.groupName()
+
+    val imageUrlSquare: String
+        get() = libXMTPGroup.groupImageUrlSquare()
+
+    val description: String
+        get() = libXMTPGroup.groupDescription()
+
+    val pinnedFrameUrl: String
+        get() = libXMTPGroup.groupPinnedFrameUrl()
```
nplasterer, line 49: Should we expose some of these things in the dm?

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -433,73 +500,303 @@ data class Conversations(
...
+                        val matchingConversations = conversationsByTopic.filter {
+                            client.inboxIdFromAddress(it.value.peerAddress)?.lowercase() ==
+                                    dm.peerInboxId().lowercase()
+                        }
+
+                        // Only send the DM if no V2 DM exists
+                        if (matchingConversations.none { it.value.version == Conversation.Version.V2 }) {
+                            trySend(Conversation.Dm(dm)).isSuccess
+                        }
+                    }
```
nplasterer, line 584: So this is where I think I need the most thinking. Feels like it might be time to keep some sort of map of V2 conversations in libxmtp so we know if the V2 conversation equivalent exists or not. If it exists in V2 then we don't want to send dms for this V3 conversation unless the client does not have a V2 identity. I think the only performant way to list all conversations to do this lookup is going to be if we have a database of the V2 conversations in libxmtp. We don't need all the messages just invites. Thoughts @richardhuaaa?

library/src/main/java/org/xmtp/android/library/Dm.kt
richardhuaaa: It might be easiest for now if these things were not modifiable in DM's, both because they're not available in v2, and because they're harder to carry over between duplicate v3 DM's

library/src/main/java/org/xmtp/android/library/Conversations.kt
richardhuaaa: I think it might make sense to add a `last_message_ns` field to groups in libxmtp, which would be useful for multiple purposes

library/src/main/java/org/xmtp/android/library/Conversations.kt
richardhuaaa: It makes sense to start storing v2 conversations in the libxmtp DB. We eventually need to back up historical v2 conversations anyway once we stop supporting v2. But let's sync up on the other behavior described here!

> If it exists in V2 then we don't want to send dms for this V3 conversation unless the client does not have a V2 identity

I think we would want to send in both, but let's sync up

richardhuaaa (commented):
Superficial first pass - I think I'm having trouble understanding class boundaries/what each file is intended to be responsible for, would love to do a quick sync-up/rundown!

In general, it seems we have a matrix of underlying communication channels on two dimensions:
1. V1 vs V2 vs V3
2. 1:1 vs group

And these map to various classes here. There is also a compositional hierarchy where some classes hold other classes, just want to develop a really clear intuition on the structure here.

library/src/main/java/org/xmtp/android/library/PreparedMessage.kt
```diff
@@ -14,7 +15,8 @@ data class PreparedMessage(
     // The first envelope should send the message to the conversation itself.
     // Any more are for required intros/invites etc.
     // A client can just publish these when it has connectivity.
-    val envelopes: List<Envelope>
+    val envelopes: List<Envelope>,
+    val encodedContent: EncodedContent? = null
```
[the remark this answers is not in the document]
nplasterer: Correct!

neekolas (approved):

insipx (approved):

codabrink (approved):

PR #310, opened 2024-10-29
Improve DM performance

nplasterer:
Fetch peer inbox id from the database.

## Release Notes

- **New Features**
	- Introduced a new property `peerInboxId` for easier access to the peer's inbox ID in direct messaging.
  
- **Improvements**
	- Enhanced error handling in message streaming functions, providing better error logging and diagnostics.
	- Simplified access to the `peerInboxId` by changing method calls to property access.

- **Version Update**
	- Updated to version `257063e9` with a new release date of `2024-10-29`.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -49,6 +49,7 @@ import uniffi.xmtpv3.FfiListMessagesOptions
 import uniffi.xmtpv3.FfiMessage
 import uniffi.xmtpv3.FfiMessageCallback
 import uniffi.xmtpv3.FfiPermissionPolicySet
+import uniffi.xmtpv3.FfiSubscribeException
 import uniffi.xmtpv3.FfiV2SubscribeRequest
 import uniffi.xmtpv3.FfiV2Subscription
 import uniffi.xmtpv3.FfiV2SubscriptionCallback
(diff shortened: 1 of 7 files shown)


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -572,6 +577,10 @@ data class Conversations(
             override fun onConversation(conversation: FfiConversation) {
                 trySend(Conversation.Group(Group(client, conversation)))
             }
+
+            override fun onError(error: FfiSubscribeException) {
+
+            }
```
nplasterer, line 583: @insipx @rygine how are you all handling the errors on the web? Should I be throwing here? Or just ignoring the error. @alexrisch what would you prefer on converse? If the stream has issues would you like me to throw something you can catch or just silently error?

nplasterer: My inclination is to just Log an error and do nothing.

insipx: Ignoring the error would be the same behavior as previous. For MessageKit we're letting the user decide how to handle the error (this allows user to restart the stream if the server restarts for instance). we haven't really discussed a strategy for web yet. It would be good to have a consistent strategy across the SDKs, though. Probably OK that MessageKit is a little different since that's for bots

codabrink (approved):

PR #314, opened 2024-11-03
[V3] Remove contacts and rename to preferences

nplasterer:
Clean up the contacts part of the sdk to be clearer about it being preferences.


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -316,11 +314,11 @@ data class Conversations(
...
-        val contacts = client.contacts
-        contacts.refreshConsentList()
-        if (contacts.consentList.state(peerAddress) == ConsentState.UNKNOWN) {
-            contacts.allow(listOf(peerAddress))
-        }
+//        val contacts = client.preferences
+//        contacts.refreshConsentList()
+//        if (contacts.consentList.state(peerAddress) == ConsentState.UNKNOWN) {
+//            contacts.allow(listOf(peerAddress))
+//        }
```
codabrink, line 321: Do we want to remove these lines?

library/src/main/java/org/xmtp/android/library/PrivatePreferences.kt
```diff
@@ -0,0 +1,142 @@
...
+    ALLOWED,
+    DENIED,
+    UNKNOWN;
+
+    companion object {
+        fun toFfiConsentState(option: ConsentState): FfiConsentState {
+            return when (option) {
+                ALLOWED -> FfiConsentState.ALLOWED
+                DENIED -> FfiConsentState.DENIED
+                else -> FfiConsentState.UNKNOWN
```
codabrink, line 17: I ran into this in Rust too when matching between the proto and native consent state and was tempted to do `_ -> UNKNOWN`, but it's probably better to be explicit and do `UNKNOWN -> UNKNOWN` so this isn't a footgun in the future when we might add another consent state and the compiler happily matches it to `UNKNOWN`.

library/src/main/java/org/xmtp/android/library/PrivatePreferences.kt
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+
+        fun fromFfiConsentState(option: FfiConsentState): ConsentState {
+            return when (option) {
+                FfiConsentState.ALLOWED -> ALLOWED
+                FfiConsentState.DENIED -> DENIED
+                else -> UNKNOWN
```
codabrink, line 25: Same here.

codabrink (approved):
Just the comments, otherwise great!

library/src/main/java/org/xmtp/android/library/Conversations.kt
nplasterer: Good call this should be handled in libxmtp now.

PR #315, opened 2024-11-03
[V3] Remove the ability to create a V2 client

nplasterer:
Major clean up removing all the V2 files from the code


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -111,85 +73,6 @@ class Client() {
...
-                    EnvelopeBuilder.buildFromTopic(
-                        topic = Topic.userPrivateStoreKeyBundle(v1Key.walletAddress),
-                        timestamp = Date(),
-                        message = encryptedKeys.toByteArray(),
-                    ),
-                ),
-            )
-        }
-
-        suspend fun canMessage(peerAddress: String, options: ClientOptions? = null): Boolean {
```
nplasterer, line 158: Maybe it would be helpful for people to still be able to use this to see if someone is on V2 if canMessageV3 returns false? They can't do much with that information tho since they can't actually message or create in V2.

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -134,7 +133,7 @@ sealed class Conversation {
         limit: Int? = null,
         before: Date? = null,
         after: Date? = null,
-        direction: PagingInfoSortDirection = MessageApiOuterClass.SortDirection.SORT_DIRECTION_DESCENDING,
```
codabrink, line 137: Wow, that's verbose

library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -1,94 +1,94 @@
...
+//                Base64.encodeToString(digest, Base64.NO_WRAP)
+//            }
+//        }
+//    }
+//
+//    private fun sha256(input: ByteArray): ByteArray {
+//        val digest = MessageDigest.getInstance("SHA-256")
+//        return digest.digest(input)
+//    }
+//}
```
codabrink, line 94: Do we want to just delete this?

codabrink (approved):

PR #316, opened 2024-11-03
[V3] Fix up all the tests

nplasterer:
Get a good test suite going


library/src/androidTest/java/org/xmtp/android/library/FramesTest.kt
```diff
@@ -1,71 +1,71 @@
...
+//        val mediaUrl = framesClient.proxy.mediaUrl(url = imageUrl!!)
+//
+//        val url = URL(mediaUrl)
+//        val connection = url.openConnection() as HttpURLConnection
+//        connection.requestMethod = "GET"
+//        val responseCode = connection.responseCode
+//        assertEquals(responseCode, 200)
+//        assertEquals(connection.contentType, "image/png")
+//    }
+//}
```
codabrink, line 71: Do we want to just delete this?

PR #317, opened 2024-11-03
[V3] Update example app

nplasterer:
Get the example showing a whole V3 only experience

### DIFF
--- a/example/src/main/java/org/xmtp/android/example/ClientManager.kt
+++ b/example/src/main/java/org/xmtp/android/example/ClientManager.kt
@@ -12,7 +12,6 @@ import org.xmtp.android.library.Client
 import org.xmtp.android.library.ClientOptions
 import org.xmtp.android.library.XMTPEnvironment
 import org.xmtp.android.library.codecs.GroupUpdatedCodec
-import org.xmtp.android.library.messages.PrivateKeyBundleV1Builder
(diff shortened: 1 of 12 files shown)


example/src/main/java/org/xmtp/android/example/ClientManager.kt
```diff
@@ -32,9 +31,8 @@ object ClientManager {
                 appVersion = "XMTPAndroidExample/v1.0.0",
                 isSecure = true
             ),
-            enableV3 = true,
             appContext = appContext,
-            dbEncryptionKey = encryptionKey
+            dbEncryptionKey = encryptionKey!!
```
codabrink, line 35: Looks like this is effectively an `.unwrap()`? What would you think if we changed the above assignment on lines 22-26? Claude helped me out here. Ignore if you want.

```
val encryptionKey = keyUtil.retrieveKey(address)?.takeUnless { it.isEmpty() }
        ?: SecureRandom().generateSeed(32).also { keyUtil.storeKey(address, it) }
```

codabrink (approved):

example/src/main/java/org/xmtp/android/example/ClientManager.kt
nplasterer: Thats some witchy Kotlin right there! 😄 Love it.

PR #320, opened 2024-11-06
V3 Frames support

nplasterer:
Support for V3 frames


library/src/main/java/org/xmtp/android/library/frames/FramesClient.kt
```diff
@@ -1,94 +1,90 @@
...
+        val trustedData = FramePostTrustedData(signedAction)
+
+        return FramePostPayload("xmtp@$PROTOCOL_VERSION", untrustedData, trustedData)
+    }
+
+    private suspend fun signDigest(digest: ByteArray): Signature {
+        val signedPrivateKey = xmtpClient.keys.identityKey
+        val privateKey = PrivateKeyBuilder.buildFromSignedPrivateKey(signedPrivateKey)
+        return PrivateKeyBuilder(privateKey).sign(digest)
+    }
```
nplasterer, line 52: Since there is no more key bundles how might we do the signing in V3? @neekolas

neekolas: We would need to sign with the installation keys (and include the pub key in the payload so it could be verified). 

This does change how the verification works. The verifier would need to pull down the identity update history for the right inbox ID to make sure the public key is registered. 

That’s going to require some work on the Node SDK side, since the verifier isn’t necessarily a full client with a database. We would want some static methods to help support workflows like this.

neekolas: For security reasons we are going to want a different context/label on the Frames signatures than what we use for messages or identity updates. That’ll help prevent replays.

nplasterer: Sounds like there might be some libxmtp work here. Probably makes sense to write up a tech doc or something. @mchenani this might be a great next task for you if you're up for it.

mchenani (approved):

PR #333, opened 2024-11-20
Consent Streaming

nplasterer:
Consent auto syncs now and streams.

### DIFF
--- a/library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
@@ -211,14 +211,14 @@ class ConversationsTest {
             boDm?.sync()
             alixClient.conversations.sync()
             alixClient2.conversations.sync()
-            alixClient2.syncConsent()
+            alixClient2.preferences.syncConsent()
             alixClient.conversations.syncAllConversations()
             Thread.sleep(2000)
             alixClient2.conversations.syncAllConversations()
             Thread.sleep(2000)
             val dm2 = alixClient2.findConversation(dm.id)!!
             assertEquals(ConsentState.DENIED, dm2.consentState())
-            alixClient2.preferences.consentList.setConsentState(
+            alixClient2.preferences.setConsentState(
                 listOf(
                     ConsentListEntry(
                         dm2.id,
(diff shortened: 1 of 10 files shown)


library/src/main/java/org/xmtp/android/library/PrivatePreferences.kt
```diff
@@ -85,10 +92,32 @@ data class ConsentListEntry(
...
     private val ffiClient: FfiXmtpClient,
 ) {
+    suspend fun syncConsent() {
+        ffiClient.sendSyncRequest(FfiDeviceSyncKind.CONSENT)
+    }
+
+    suspend fun streamConsent(): Flow<ConsentListEntry> = callbackFlow {
+        val consentCallback = object : FfiConsentCallback {
+            override fun onConsentUpdate(consent: List<FfiConsent>) {
+                consent.forEach {
```
cameronvoell, line 106:
```suggestion
                consent.iterator().forEach {
```

library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
```diff
@@ -228,10 +228,76 @@ class ConversationsTest {
...
+            val dm3 = alixClient2.conversations.newConversation(caro.walletAddress)
+            dm3.updateConsentState(ConsentState.DENIED)
+            alixClient.conversations.sync()
+            alixClient2.conversations.sync()
+            alixClient.conversations.syncAllConversations()
+            alixClient2.conversations.syncAllConversations()
+        }
+
+        Thread.sleep(2000)
+        assertEquals(3, consent.size)
```
cameronvoell, line 299: Checking if I understand - there are 3 consent entries because when `alixClient2` creates the `dm3` group with caro, it automatically makes an entry of ALLOWED, and the other two entries are making the `alix2Group` DENIED, and setting the `dm3` group DENIED. Is that right?

cameronvoell (approved):

library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
nplasterer: Correct!

PR #334, opened 2024-11-21
Signature verification

nplasterer:
Adds the ability to verify if a signature was signed or not by a key

### DIFF
--- a/library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
@@ -413,7 +413,7 @@ class ClientTest {
             false
         )
     }
-    
+
     @Test
     fun testAddAccounts() {
         val fixtures = fixtures()
--- a/library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
@@ -354,28 +354,32 @@ class ClientTest {
                 "Testing",
                 signature,
                 alixInstallationId
-            ), true
+            ),
+            true
         )
         assertEquals(
             fixtures.alixClient.verifySignatureWithInstallationId(
                 "Not Testing",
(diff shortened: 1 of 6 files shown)


library/src/androidTest/java/org/xmtp/android/library/ClientTest.kt
```diff
@@ -331,7 +331,80 @@ class ClientTest {
...
+                fixtures.boClient.installationId
+            ), false
+        )
+        assertEquals(
+            fixtures.boClient.verifySignatureWithInstallationId(
+                "Testing",
+                signature,
+                alixInstallationId
+            ), true
+        )
```
nplasterer, line 379: This feels weird that it's true? Should Bo really be able to verify if a signature was made by alix's installationId?

codabrink (approved):
Looks good!

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -212,6 +212,24 @@ class Client() {
...
+    fun verifySignature(message: String, signature: ByteArray): Boolean {
+        return try {
+            ffiClient.verifySignedWithInstallationKey(message, signature)
+            true
+        } catch (e: Exception) {
+            false
+        }
+    }
+
+    fun verifySignatureWithInstallationId(message: String, signature: ByteArray, installationId: String): Boolean {
```
codabrink, line 224: Sorry, I missed this when I approved. Can we rename this function to match with the ffi client?

codabrink: Or I can rename the fn on the client if it's confusing.

nplasterer: I think it's confusing for integrators since it's not clear that the installationId is the public key. I think it's okay for the names between the SDKs and libxmtp to differ. This isn't the first time I've strayed from the ffi naming in benefits of clarity in the sdk.

PR #344, opened 2024-12-04
Static Can Message

nplasterer:
Adds the ability to check if a address is on the V3 network without creating a client.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -77,7 +77,11 @@ class Client() {
             codecRegistry.register(codec = codec)
         }
 
-        suspend fun canMessage(accountAddresses: List<String>, appContext: Context, api: ClientOptions.Api): Map<String, Boolean> {
+        suspend fun canMessage(
+            accountAddresses: List<String>,
+            appContext: Context,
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -76,6 +76,35 @@ class Client() {
...
+                accountAddress = accountAddress.lowercase(),
+                inboxId = inboxId,
+                nonce = 0.toULong(),
+                legacySignedPrivateKeyProto = null,
+                historySyncUrl = null
+            )
+
+            val result = ffiClient.canMessage(accountAddresses)
+            ffiClient.releaseDbConnection()
+            File(dbPath).delete()
```
nplasterer, line 104: @rygine I think we will want to delete the db that is created with this temp client? Maybe something to add to the js sdk.

insipx (approved):

cameronvoell (approved):

library/src/main/java/org/xmtp/android/library/Client.kt
rygine: yep! i recently added the ability to create a browser/node clients without a DB or encryption key for this purpose. just need to update the SDKs. great reminder, thanks!

PR #346, opened 2024-12-05
Re-Enable History Sync

nplasterer:
Adds additional testing and re-enables the history sync functionality.


library/src/androidTest/java/org/xmtp/android/library/HistorySyncTest.kt
```diff
@@ -0,0 +1,173 @@
...
+        bo = fixtures.bo
+        caroWallet = fixtures.caroAccount
+        caro = fixtures.caro
+
+        alixClient = fixtures.alixClient
+        boClient = fixtures.boClient
+        caroClient = fixtures.caroClient
+    }
+    @Test
+    fun testSyncConsent() {
```
nplasterer, line 46: Two other tests that we should write is calling sync messages and automatic syncing for consent and messages.

codabrink (approved):

PR #352, opened 2024-12-16
Libxmtp fork fix dont overwrite intent state

cameronvoell:
See https://github.com/xmtp/libxmtp/pull/1416

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -129,7 +129,8 @@ data class Conversations(
 
     // Sync all new and existing conversations data from the network
     suspend fun syncAllConversations(): UInt {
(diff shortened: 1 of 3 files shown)


library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -129,7 +129,8 @@ data class Conversations(
 
     // Sync all new and existing conversations data from the network
     suspend fun syncAllConversations(): UInt {
-        return ffiConversations.syncAllConversations()
+        // TODO: add consentState here
+        return ffiConversations.syncAllConversations(null)
```
cameronvoell, line 133: @nplasterer per conversation we'll add this in a follow up. or let me know if I misunderstood

nplasterer (approved):

PR #353, opened 2024-12-17
Sync All Conversations Consent Filtering 

nplasterer:
Adds the ability to filter sync all conversations by consent state
This also brings in an additional fix for forked groups


mchenani (approved):

cameronvoell (approved):
looking good 👍

PR #354, opened 2024-12-18
Speed up build client performance

nplasterer:
Much much faster now by passing in the apiClient.

Message | Duration
-- | --
Created a client | 1.61s
Built a client | 0.886s
Built a client with inboxId | 0.41s
Built a client with inboxId and apiClient | 0.111s


cameronvoell (approved):
lint warning to remove an extra break line

awesome performance improvement!

PR #360, opened 2025-01-06
Message refactor and cleanup

nplasterer:
Remove DecodedMessages in favor of just Messages and cleans up some incorrect imports.


library/src/main/java/org/xmtp/android/library/libxmtp/Message.kt
```diff
@@ -45,33 +44,34 @@ data class Message(val client: Client, private val libXMTPMessage: FfiMessage) {
...
-        } catch (e: Exception) {
-            throw XMTPException("Error decoding message", e)
+    val topic: String
+        get() = Topic.groupMessage(convoId).description
+
+    val encodedContent: Content.EncodedContent
+        get() = EncodedContent.parseFrom(libXMTPMessage.content)
+
+    fun <T> content(): T? =
+        encodedContent.decoded()
```
cameronvoell, line 54: I'm thinking the decoding should maybe happen in the constructor of the Message instead of in the content() function. That way, that way if you call content() on a message multiple times it doesnt repeat the decoding logic. wdyt?

cameronvoell (approved):
Like the cleanup - one suggestion about decoding in Message constructor instead of the content function

library/src/main/java/org/xmtp/android/library/libxmtp/Message.kt
nplasterer: That makes sense!

PR #361, opened 2025-01-08
Conversations allTopics for dm stitching

nplasterer:
Expose a method that subscribes all topics that includes duplicate dms.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Conversations.kt
+++ b/library/src/main/java/org/xmtp/android/library/Conversations.kt
@@ -299,4 +299,17 @@ data class Conversations(
         }
         return hmacKeysResponse.build()
     }
+
+    fun allTopics(): List<String> {
+        val conversations = ffiConversations.list(
+            FfiListConversationsOptions(
+                null,
+                null,
+                null,
+                null,
+                includeDuplicateDms = true
+            )
+        )
+        return conversations.map { Topic.groupMessage(it.conversation().id().toHex()).description }
+    }
 }
--- a/example/src/main/java/org/xmtp/android/example/MainViewModel.kt
+++ b/example/src/main/java/org/xmtp/android/example/MainViewModel.kt
@@ -45,11 +45,11 @@ class MainViewModel : ViewModel() {
         viewModelScope.launch(Dispatchers.IO) {
             val listItems = mutableListOf<MainListItem>()
             try {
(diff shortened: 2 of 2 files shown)


library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
```diff
@@ -208,4 +210,53 @@ class ConversationsTest {
...
+            boClient.conversations.syncAllConversations()
+        }
+
+        val topics = boClient.conversations.allTopics()
+        val conversations = runBlocking { boClient.conversations.list() }
+        val hmacKeys = boClient.conversations.getHmacKeys()
+
+        assertEquals(topics.size, 2)
+        assertEquals(conversations.size, 1)
+        assertEquals(conversations.first().id, dm1.id)
```
nplasterer, line 255: @codabrink what call needs to happen for eriClient1 and eriClient2 to both see the same dm instead of two different ones?

codabrink: Someone needs to send a message. That will cause the "primary DM" to be chosen.

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -299,4 +299,17 @@ data class Conversations(
...
     }
+
+    fun allTopics(): List<String> {
+        val conversations = ffiConversations.list(
+            FfiListConversationsOptions(
+                null,
+                null,
+                null,
+                null,
+                includeDuplicateDms = true
```
codabrink, line 310: :+1:

codabrink (approved):

nplasterer:
Going to wait to merge this till I have a few additional helpful functions https://github.com/xmtp/libxmtp/issues/1499

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -299,4 +299,17 @@ data class Conversations(
         }
         return hmacKeysResponse.build()
     }
+
+    fun allTopics(): List<String> {
```
richardhuaaa, line 303: Could we call this something like `getPushTopics()`, to kill the possibility that integrators misuse this?

cameronvoell (approved):

PR #369, opened 2025-01-15
Clean up redundant consent code

nplasterer:
Removes consent set on send message as it's handled in libxmtp.


cameronvoell (approved):
just noting here that removing this did not fix the db is locked errors on send message. still good to clean this up 👍

PR #371, opened 2025-01-17
Consent filtering by array

nplasterer:
Add ability to filter and sync by multiple consents
And new inboxId methods for performance
Last readable message
Disables message history -- while we look through database locking issues


cameronvoell (approved):
looks good!

PR #374, opened 2025-02-03
Remove pinned frame url

nplasterer:
Dependent on PR https://github.com/xmtp/libxmtp/pull/1534 landing first

We could do this in a less breaking way and just set the pinned frame always to an empty string if we want to merge this before libxmtp changes land but this would be the cleanest.

- Remove pinned frame url
- Defaults History Sync back on


codabrink (approved):

mchenani (approved):

cameronvoell:
Pausing merging while we look into https://github.com/xmtp/libxmtp/issues/1589

PR #375, opened 2025-02-04
Disappearing messages

nplasterer:
Adds the ability to update, set, and get expiration settings for messages on a DM or Group.


library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -337,6 +352,14 @@ class Group(
         )
     }
 
+    suspend fun updateMessageExpirationPermission(newPermissionOption: PermissionOption) {
+        return libXMTPGroup.updatePermissionPolicy(
+            FfiPermissionUpdateType.UPDATE_METADATA,
+            PermissionOption.toFfiPermissionPolicy(newPermissionOption),
+            FfiMetadataField.MESSAGE_EXPIRATION
```
nplasterer, line 359: Need this from libxmtp

library/src/main/java/org/xmtp/android/library/Conversation.kt
```diff
@@ -36,6 +38,14 @@ sealed class Conversation {
             }
         }
 
+    val messageExpiration: MessageExpiration
```
mchenani, line 41: suggestion: we used message disappearingSettings everywhere, I would suggest let's have the same naming here too. wdyt?

library/src/main/java/org/xmtp/android/library/Group.kt
mchenani: Since we set the messageDisappearing setting always be admin, I don't think we need this, do we?

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -1022,4 +1023,55 @@ class GroupTest {
...
+        runBlocking {
+            boClient.conversations.syncAllConversations()
+            alixClient.conversations.syncAllConversations()
+        }
+        message = boClient.findMessage(messageId)
+        alixMessage = alixClient.findMessage(messageId)
+        assert(message == null)
+        assertEquals(runBlocking { boGroup.messages() }.size, 0)
+        assert(alixMessage == null)
+        assertEquals(runBlocking { alixGroup!!.messages() }.size, 0)
```
nplasterer, line 1057: @mchenani my messages aren't disappearing like I would expect is there something I need to do to trigger the worker to run?

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -1022,4 +1023,55 @@ class GroupTest {
...
+
+        // Can disable message expiration
+        val messageDisappearingSettings = MessageDisappearingSettings(0L, 0L)
+        runBlocking {
+            alixGroup!!.updateMessageExpiration(messageDisappearingSettings)
```
nplasterer, line 1062: Is this how you turn off message disappearing settings? Seems like the update method in libxmtp should be nullable instead?

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
mchenani: did you start the worker? and you need to wait for a second for the worker, the worker clean up the messages every second.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
mchenani: We can't set the values to null, because if the value set to null, we can't detect it in the metaDataUpdate process.
but for removing the settings, you can call `remove_conversation_message_disappearing_settings`. that would set the default values for you.

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -1022,4 +1023,54 @@ class GroupTest {
...
+        boGroup.send("howdy2")
+        alixGroup!!.send("howdy howdy")
+
+        assertEquals(boGroup.messages().size, 4)
+        assertEquals(alixGroup.messages().size, 4)
+        boClient.conversations.syncAllConversations()
+        alixClient.conversations.syncAllConversations()
+        Thread.sleep(4000)
+
+        assertEquals(boGroup.messages().size, 2)
```
nplasterer, line 1057: @mchenani I tried to match this as closely to the rust test but it's still not working like expected?

library/src/androidTest/java/org/xmtp/android/library/DmTest.kt
```diff
@@ -400,4 +402,114 @@ class DmTest {
...
+        alixDm.send("another message after disabling")
+        boDm.sync()
+
+        Thread.sleep(1000)
+
+        // Ensure messages persist
+        assertEquals(
+            boDm.messages().size,
+            2
+        ) // disappearing settings 1, disappearing settings 2, boMessage, alixMessage
```
nplasterer, line 453: All these counts will need to be updated once we start showing group update messages in the dms. @mchenani

mchenani (approved):

library/src/main/java/org/xmtp/android/library/libxmtp/DisappearingMessageSettings.kt
```diff
@@ -0,0 +1,14 @@
+package org.xmtp.android.library.libxmtp
+
+import uniffi.xmtpv3.FfiMessageDisappearingSettings
+
+class DisappearingMessageSettings(
+    val disappearStartingAtNs: Long,
+    val disappearDurationInNs: Long,
```
cameronvoell, line 7: second var name seems confusing in my opinion. maybe `retentionDurationInNs` is clearer?

cameronvoell (approved):
On small var name suggestion, otherwise LGTM! nice tests 👌

library/src/main/java/org/xmtp/android/library/libxmtp/DisappearingMessageSettings.kt
nplasterer: What about `disappearingInNS`? Or `disappearingIntervalNs`?

PR #381, opened 2025-02-11
Make addWallet a delicate API

nplasterer:
Require a boolean to be passed to allow overriding an existing inboxId when calling `addWallet`. Also makes a new annotation for delicate Apis.

### DIFF
--- a/library/src/main/java/org/xmtp/android/library/Client.kt
+++ b/library/src/main/java/org/xmtp/android/library/Client.kt
@@ -259,10 +259,11 @@ class Client() {
     }
 
     @DelicateApi("This function is delicate and should be used with caution. Adding a wallet already associated with an inboxId will cause the wallet to loose access to that inbox. See: inboxIdFromAddress(address)")
-    suspend fun addAccount(newAccount: SigningKey, changeInboxId: Boolean = false) {
-        val inboxId: String? = if (!changeInboxId) inboxIdFromAddress(newAccount.address) else null
+    suspend fun addAccount(newAccount: SigningKey, allowReassignInboxId: Boolean = false) {
+        val inboxId: String? =
+            if (!allowReassignInboxId) inboxIdFromAddress(newAccount.address) else null
 
-        if (changeInboxId || inboxId.isNullOrBlank()) {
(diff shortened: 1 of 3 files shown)


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -258,11 +258,17 @@ class Client() {
         ffiClient.applySignatureRequest(signatureRequest)
     }
 
-    suspend fun addAccount(newAccount: SigningKey) {
-        val signatureRequest =
-            ffiClient.addWallet(newAccount.address.lowercase())
-        handleSignature(signatureRequest, newAccount)
-        ffiClient.applySignatureRequest(signatureRequest)
+    @DelicateApi("This function is delicate and should be used with caution. Adding a wallet already associated with an inboxId will cause the wallet to loose access to that inbox. See: inboxIdFromAddress(address)")
```
neekolas, line 261:
```suggestion
    @DelicateApi("This function is delicate and should be used with caution. Adding a wallet already associated with an inboxId will cause the wallet to lose access to that inbox. See: inboxIdFromAddress(address)")
```

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -258,11 +258,17 @@ class Client() {
...
+    suspend fun addAccount(newAccount: SigningKey, changeInboxId: Boolean = false) {
```
neekolas, line 262: What about something with "allow" in it? It will only change the inbox ID if the new account already has one, which is not going to be true most of the time.

Something like `allowReassignInboxId`?

cameronvoell (approved):
Looks good - I would maybe add a test too if nothing else just to help document how it is expected to work. But I think this is a good protection against integrators accidentally doing this. 👍

nplasterer:
@jhaaaa we should probably also add some extra warnings and notes to the docs about adding wallets. 🙏

jhaaaa:
> @jhaaaa we should probably also add some extra warnings and notes to the docs about adding wallets. 🙏

Done with the V3 doc release - thank you! https://docs.xmtp.org/inboxes/manage-inboxes#add-an-identity-to-an-inbox

PR #384, opened 2025-02-20
Performance testing

nplasterer:
- Performance Tests
- Streaming Tests
- Binding bump
- Function move from client to conversations to match other SDKs

| Operation                                   |   Time (ms) (Dev) |   Time (ms) (Local) |
|:--------------------------------------------|------------------:|--------------------:|
| created a DM                                |              1001 |                 116 |
| sendGmTime                                  |               305 |                  37 |
| createGroupTime                             |              1021 |                  68 |
| sendGmInGroupTime                           |               304 |                  21 |


library/src/androidTest/java/org/xmtp/android/library/ConversationsTest.kt
```diff
@@ -253,4 +262,101 @@ class ConversationsTest {
...
+
+//        val boSpamJob = launch(Dispatchers.IO) {
+//            println("Bo is sending spam groups..")
+//            repeat(10) {
+//                val spamMessage = "Bo Message $it"
+//                val group = boClient.conversations.newGroup(listOf(caroClient.address))
+//                group.send(spamMessage)
+//                println("Bo spam: $spamMessage")
+//            }
+//        }
```
nplasterer, line 330: @insipx I added this test and it passes about 80% of the time. Occasionally misses the first sent message but otherwise is pretty reliable. 
However when I add in this group creation it doesn't seem to stream messages from new groups? If I get added to a group I shouldn't have to resubscribe to `streamAllMessages` right? Or is this expected behavior.

insipx: you should not have to re-subscribe, the stream should pick up on the new group and add it. in `xmtp_mls` we have Eve create a group and send a message to test this out, if its not working then there's a bug somewhere. This test seems more involved then the libxmtp one, so ill try adding in more things to the test in xmtp_mls and see if i can reproduce that behavior

codabrink (approved):

PR #389, opened 2025-02-28
Passkey Read

nplasterer:
- [x] Make sure addWallet is blocking reassignment for the ffi function as well
- [x] Adds new identity for all places where addresses were used
- [x] Changes `addMember`, `removeMember`, `createGroup`, and `newConversation` to default to `InboxId`
- [x] Enforces type on inboxId so address can't accidentally be passed.


library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -287,10 +292,14 @@ class Client(
     @DelicateApi("This function is delicate and should be used with caution. Adding a wallet already associated with an inboxId will cause the wallet to lose access to that inbox. See: inboxIdFromAddress(address)")
     suspend fun addAccount(newAccount: SigningKey, allowReassignInboxId: Boolean = false) {
```
nplasterer, line 293: @codabrink I'm really struggling with this piece in particular. Right now we just have one interface for signers. Either we need a RootSigner and additional Signer. 🤔 That may be okay. Let me try that.

library/src/main/java/org/xmtp/android/library/Client.kt
```diff
@@ -39,15 +41,16 @@ data class ClientOptions(
     )
 }
 
+@JvmInline
+value class InboxId(val value: String)
```
nplasterer, line 45: I made this a value class so that it would error if people pass an address instead of an inboxId into certain methods. If you pass an address instead of a inboxId you get this error: https://github.com/xmtp/libxmtp/issues/1693

There are less aggressive ways to do typing like a typealias... but it wouldn't give clear in line coding warnings which is what I think is necessary here with this level of a change. Open to thoughts though.

library/src/main/java/org/xmtp/android/library/Group.kt
```diff
@@ -213,49 +214,49 @@ class Group(
...
+    suspend fun creatorInboxId(): InboxId {
+        return InboxId(metadata().creatorInboxId())
     }
 
     suspend fun isCreator(): Boolean {
-        return metadata().creatorInboxId() == client.inboxId
+        return metadata().creatorInboxId() == client.inboxId.value
     }
 
-    suspend fun addMembers(addresses: List<String>) {
```
nplasterer, line 232: Big breaking change here is `addMembers` now takes a `inboxId` instead of a address... but because they are both strings it will blow up only after it's been hit in the code.

library/src/main/java/org/xmtp/android/library/Conversations.kt
```diff
@@ -318,23 +301,23 @@ data class Conversations(
         return Dm(client, dmConversation)
     }
 
-    suspend fun newConversationWithInboxId(
-        peerInboxId: String,
+    suspend fun newConversation(
+        peerInboxId: InboxId,
```
nplasterer, line 305: Big breaking change here is newConversation and new Group now takes a inboxId instead of a address... but because they are both strings it will blow up only after it's been hit in the code.

cameronvoell (approved):
looking great!

cameronvoell (commented):
Noticed something odd, `testCannotStartGroupWithSelf` started failing in this branch, looks good on `main`

looks like no error thrown when you start a group with yourself 🤔

GroupTest.kt:483

```kotlin
@Test
    fun testCannotStartGroupWithSelf() {
        assertThrows("Recipient is sender", XMTPException::class.java) {
            runBlocking { boClient.conversations.newGroup(listOf(boClient.inboxId)) }
        }
    }
```

nplasterer:
Ahh ya I thought I fixed the self dms error. But honestly this should probably be enforced in libxmtp. I'll look into making the fix in the lower level.

library/src/main/java/org/xmtp/android/library/Client.kt
mchenani: I agree too, also I would say we need to do it more and more, also we're doing it in rust more often.

PR #392, opened 2025-03-06
Should Push for Push Notifications

nplasterer:
Adds the should_push field to messages so that push notification servers can respect them accordingly.


library/src/main/java/xmtpv3.kt
nplasterer, line 1: Big diff change because of the change to `cargo NDK` but all tests are green.

mchenani (approved):

PR #394, opened 2025-03-10
Update Signer to be more universal

nplasterer:
Lets make the signer more universal so it can handle more signature types long term.


library/src/main/java/libxmtp-version.txt
```diff
@@ -1,3 +1,3 @@
-Version: c2b99064
+Version: e1bbb348
 Branch: main
-Date: 2025-03-06 17:48:33 +0000
+Date: 2025-03-08 01:05:24 +0000
```
nplasterer, line 3: This is the release candidate from Friday.

nplasterer: But sounds like we're going to make another one today with the web fix so we can bump again.

library/src/main/java/org/xmtp/android/library/libxmtp/GroupMembershipResult.kt
```diff
@@ -0,0 +1,15 @@
+package uniffi.xmtpv3.org.xmtp.android.library.libxmtp
+
+import org.xmtp.android.library.InboxId
+import org.xmtp.android.library.toHex
+import uniffi.xmtpv3.FfiUpdateGroupMembershipResult
+
+class GroupMembershipResult(private val ffiUpdateGroupMembershipResult: FfiUpdateGroupMembershipResult) {
+    val addedMembers: Map<InboxId, ULong>
```
nplasterer, line 8: @mchenani Just confirming the string in `addMembers` is the `InboxId` 👍

cameronvoell (approved):
🚀

PR #405, opened 2025-03-27
Hot fix for key packages

nplasterer:
## Update XMTP Native Libraries and Version Information
Updated native `libuniffi_xmtpv3.so` libraries for all architectures (arm64-v8a, armeabi-v7a, x86, x86_64) and version tracking information in version.txt

### 📍Where to Start
Start with [libxmtp-version.txt](https://github.com/xmtp/xmtp-android/pull/405/files#diff-f6791b18883003e6e2f4bfe862883cea4ff5c62c2ab636c0e7e47be4cda8bf09R0) to verify version changes, then examine the binary changes in the `libuniffi_xmtpv3.so` files for each architecture in the `jniLibs` directory

----

_[Macroscope](https://app.macroscope.com) summarized ddaa436._ (Automatic summaries will resume when PR exits draft mode or review begins).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release XMTP Android Package
 on:
   push:
     branches:
-      - release
+      - hot-fix-key-package
 
 jobs:
   library:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,21 +22,20 @@ jobs:
         uses: gradle/actions/setup-gradle@v3
       - name: Run build with Gradle Wrapper
         run: ./gradlew build
-      - name: Bump version and push tag
-        id: tag_version
-        uses: mathieudutour/github-tag-action@v6.1
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -22,21 +22,20 @@ jobs:
...
-      - name: Bump version and push tag
-        id: tag_version
-        uses: mathieudutour/github-tag-action@v6.1
-        with:
-          release_branches: "release"
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Create a GitHub release
+      - name: Tag version 3.1.0
+        run: |
+          git tag 3.1.0
```
nplasterer, line 27: @cameronvoell thoughts on versioning here? I could also be down with `3.0.29-patch1` or something?

nplasterer: Also going to have to update the release script to pull from this branch instead of main to make this happen.

PR #421, opened 2025-05-07
Optimistically create groups

nplasterer:
Create a group offline to be synced later.

### DIFF
--- a/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
+++ b/library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
@@ -601,7 +601,7 @@ class GroupTest {
             },
             2
         )
-        assertEquals(runBlocking { boClient.conversations.listGroups().size }, 2)
+        assertEquals(runBlocking { boClient.conversations.listGroups().size }, 1)
     }
 
     @Test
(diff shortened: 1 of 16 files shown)


library/src/main/java/org/xmtp/android/library/Dm.kt
```diff
@@ -41,7 +41,7 @@ class Dm(
         get() = Date(libXMTPGroup.createdAtNs() / 1_000_000)
 
     val peerInboxId: InboxId
-        get() = libXMTPGroup.dmPeerInboxId()
+        get() = libXMTPGroup.dmPeerInboxId() ?: ""
```
nplasterer, line 44: @cameronvoell this is nullable now. What do you think would be the best solution? It would be a breaking change to make it `InboxId?` What case would it be empty?

library/src/androidTest/java/org/xmtp/android/library/GroupTest.kt
```diff
@@ -1028,7 +1028,7 @@ class GroupTest {
 
         assertEquals(runBlocking { alixGroup.messages() }.size, 1)
         assertEquals(runBlocking { alixGroup2.messages() }.size, 1)
-        assertEquals(numGroups, 3u)
+        assertEquals(numGroups, 2u) // I still think this should be 3
```
nplasterer, line 1031: @codabrink this makes me think it should be 3 still if it was syncing the sync group correctly?

nplasterer: Apparently this is expected now that syncAll does not sync the sync Group.

cameronvoell (approved):

PR #422, opened 2025-05-09
Bump to latest bindings

nplasterer:
Gets on the latest bindings and makes a new RC


library/src/androidTest/java/org/xmtp/android/library/HistorySyncTest.kt
```diff
@@ -134,7 +134,7 @@ class HistorySyncTest {
         }
 
         Thread.sleep(2000)
-        assertEquals(3, consent.size)
+        assertEquals(5, consent.size)
```
nplasterer, line 137: @codabrink I have no idea why this is 5 now... I'm assuming now that syncAll syncs the device sync? But 5 feels like a lot. Since I only change the consent 3 times...

rygine (approved):

